PR #5, opened 2020-12-18
Fixes call to JATON missing typecast (const char*)

gwolffblues:
Functions declared using K&R syntax, the parameters used when the function is called must match the required data types in the function definition

 This was required to get note-c to compile with clang.

https://en.cppreference.com/w/c/language/function_declaration


bsatrom:
@zfields can you field this one?

zfields (approved):

PR #6, opened 2020-12-23
Fix README links

wtfleming:
Fixes relative links that should be external.
For example, the note-arduino link was pointing to https://github.com/blues/note-c/blob/master/note-arduino instead of https://github.com/blues/note-arduino


wtfleming:
Closing, this looks like it was fixed in #10

PR #9, opened 2021-01-25
Compile time error counters in note-transaction.

sdt99:
Provide a low overhead, interface agnostic method to verify reliability.

### DIFF
--- a/n_request.c
+++ b/n_request.c
@@ -13,6 +13,24 @@
 
 #include "n_lib.h"
 
+//#define DEBUG
+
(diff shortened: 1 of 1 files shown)


m-mcgowan (approved):
LGTM! 🎉  It would be nice to have a way to retrieve a string form of the error metrics for easy printing over serial so launching a debugger isn't strictly required.

rayozzie:
Because of changes in n_request and no burning need for this, I would prefer to defer this for now.

PR #13, opened 2021-04-15
Add NoteRequestWithRetry to retry requests until success or timeout

sdt99:
I added this function to allow the Truenote host to retry certain critical requests - I'm submitting a draft PR for feedback as to whether you think this is the right approach, before I update the rest of the truenote firmware.

### DIFF
--- a/n_request.c
+++ b/n_request.c
@@ -125,6 +125,70 @@ bool NoteRequest(J *req)
 
 /**************************************************************************/
 /*!
+    @brief  Send a request to the Notecard.
+            Frees the request structure from memory after sending the request.
+            Retries the request for up to the specified timeoutSeconds if there is
+            no response, or if the response indicates an io error.
+    @param   req
+               The `J` cJSON request object.
+             timeoutSeconds
+               Upper limit for retries if there is no response, or if the 
+               response contains an io error.
+  @returns a boolean. Returns `true` if successful or `false` if an error
+            occurs, such as an out-of-memory or if an error was returned from
+            the transaction in the c_err field.
+*/
+/**************************************************************************/
+bool NoteRequestWithRetry(J *req, uint32_t timeoutSeconds)
+{
+    // Exit if null request.  This allows safe execution of the form NoteRequest(NoteNewRequest("xxx"))
+    if (req == NULL) {
+        return false;
+    }
+
+    J *rsp;
+
+    // Calculate expiry time in milliseconds
+    uint32_t expiresMs = _GetMs() + (timeoutSeconds * 1000);
+
+    while(true) {
+        // Execute the transaction
+        rsp = NoteTransaction(req);
+
+        // Loop if there is no response, or if there is an io error
+        if ( (rsp == NULL) || JContainsString(rsp, c_err, c_ioerr)) {
+
+            // Free error response
+            if (rsp != NULL) {
+                JDelete(rsp);
+            }
+        } else {
+
+            // Exit loop on non-null response without io error
+            break;
+        }
+
+        // Exit loop on timeout
+        if (_GetMs() >= expiresMs) {
+            break;
+        }
+    }
+
+    // Free the request
+    JDelete(req);
+
+    // If there is no response return false
+    if (rsp == NULL) {
+        return false;
+    }
+    // Check for a transaction error, and exit
+    bool success = JIsNullString(rsp, c_err);
+    JDelete(rsp);
+    return success;
+}
+
+/**************************************************************************/
+/*!
     @brief  Send a request to the Notecard and return the response.
             Frees the request structure from memory after sending the request.
     @param   req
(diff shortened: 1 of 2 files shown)


rayozzie (commented):
Several question/comments.
1) so you decided not to build this directly into NoteTransaction?  I understand that it's complex but i'm curious as to why
2) If you were going to expose a single call that supports retry, it should NOT be "NoteRequest" but rather it should be "NoteRequestResponse", because the latter returns "rsp" and it's easy for the caller to NoteDeleteResponse(rsp) whereas if you don't implement NoteRequestResponse there is no way to do it.  Of course, you could do both
3) if it were me I'd take the timeout as Secs, not Ms, because nobody is going to time out based on small numbers of ms
4) startms will never come back as 0.  ever.  it isn't worth checking.
5) I would not, under any circumstances, keep retrying on "any error" but rather I would only retry on transactions that contain "{io}".  It makes no semantic sense to be retrying if it doesn't contain {io}.  I would strongly urge that you to do these things: 1) change line 154 to "if (NoteResponseErrorContains(rsp, c_ioerr)) {", 2) let me change n_request.c to have behavior identical to the golang library so that {io} is inside all io errors.  (I will do that today)
6) you've got a pretty bad bug on line 169, which should be "(rsp != NULL)"

rayozzie:
I've submitted the changes to note-c so that {io} appears inside all I/O errors, so you'll need to re-pull into truenote.

sdt99:
Thanks for the feedback - a couple of things I'd like to respond to:
 if the user had not supplied a hookGetMs() function startms would come back as 0 - I was guarding against that since this is a public library.

" but rather I would only retry on transactions that contain "{io}""
The errors did not previously contain {io} - as you noted, but I was also concerned about errors of truncation, where the Notecard would reply with an error, such as 
cannot open json as object
unknown request type
cannot process request: unrecognized:
{"err":"missing 'req' field specifying the type of request"}
I was concerned that if I was too specific in which errors I chose to retry that it would not really solve the problem of i2c or serial bus related errors.
My assumption was that the user would not use this function for a request that could legitimately cause a non io-related error to be returned by the notecard.

I agree that it would be better applied to NoteRequestResponse() and with a timeout in seconds.

rayozzie:
Yes, but if they don’t supply a hookGetMs() function, absolutely nothing will work and there are no other checks anywhere else in the library for that case.  It would be odd to have a check in that method but nowhere else.

You’re right about those errors.  I’ll add {io} to them right now!  It will make the notefarm in particular much more robust.

(The problem that we’ve got for ALL devices, though, is that you’re right to protect against common I/O errors, it really does beg the question as to what you should do on *all* the initialization calls if you get some other weird error back.  Unfortunately, panics are useless to the user.)

BTW I think you should add versions for both NoteRequestResponse and NoteRequest.

From: Sean Taylor ***@***.***>
Reply-To: blues/note-c ***@***.***>
Date: Thursday, April 15, 2021 at 5:18 PM
To: blues/note-c ***@***.***>
Cc: Ray Ozzie ***@***.***>, Review requested ***@***.***>
Subject: Re: [blues/note-c] Add NoteRequestWithRetry to retry requests until success or timeout (#13)


Thanks for the feedback - a couple of things I'd like to respond to:
if the user had not supplied a hookGetMs() function startms would come back as 0 - I was guarding against that since this is a public library.

" but rather I would only retry on transactions that contain "{io}""
The errors did not previously contain {io} - as you noted, but I was also concerned about errors of truncation, where the Notecard would reply with an error, such as
cannot open json as object
unknown request type
cannot process request: unrecognized:
{"err":"missing 'req' field specifying the type of request"}
I was concerned that if I was too specific in which errors I chose to retry that it would not really solve the problem of i2c or serial bus related errors.
My assumption was that the user would not use this function for a request that could legitimately cause a non io-related error to be returned by the notecard.

I agree that it would be better applied to NoteRequestResponse() and with a timeout in seconds.

sdt99:
I have incorporated suggested changes - if this looks OK I'll make the necessary changes to the Truenote application.

A related question is whether it's worth creating parallel functions where retry might be desired:
NoteSet*(), NoteGet*() - or perhaps a configuration function that selects whether NoteSet*() and NoteGet*() should select a retry timeout or not (not NoteAdd, per your previous comments).

bsatrom (approved):
LGTM on the whole.

If we are going to consider retry variants of all functions, like NoteGet*, et all, perhaps it would make sense to do all of these as overloads to the base functions. so:

```cpp
bool NoteRequest(J *req);
bool NoteRequest(J *req, uint32_t);
bool NoteRequestResponse(J *req);
bool NoteRequestResponse(J *req, uint32_t);
```

and so on

rayozzie:
Unfortunately, we need note-c to work on all C compilers for all embedded platforms, and C (unlike C++) doesn’t support overloading.


From: Brandon Satrom ***@***.***>
Reply-To: blues/note-c ***@***.***>
Date: Wednesday, April 21, 2021 at 4:56 PM
To: blues/note-c ***@***.***>
Cc: Ray Ozzie ***@***.***>, Review requested ***@***.***>
Subject: Re: [blues/note-c] Add NoteRequestWithRetry to retry requests until success or timeout (#13)


@bsatrom approved this pull request.

LGTM on the whole.

If we are going to consider retry variants of all functions, like NoteGet*, et all, perhaps it would make sense to do all of these as overloads to the base functions. so:

bool NoteRequest(J *req);

bool NoteRequest(J *req, uint32_t);

bool NoteRequestResponse(J *req);

bool NoteRequestResponse(J *req, uint32_t);

and so on

bsatrom:
Sean reminded me of this, I forgot that C doesn’t do overloads. :/

________________________________

PR #17, opened 2022-10-18
Fix NoteRequestResponseWithRetry (Returns already freed/deleted rsp)

alvarop:
NoteRequestResponseWithRetry will free rsp but still return rsp in the case when NoteTransaction returns an error and expiresMs is reached. The caller then tries to free rsp again which results in an error.

This change makes sure to set rsp to NULL whenever it is deleted to prevent this.

As @russelldeguzman pointed out, this also affects NoteRequestWithRetry. I've added the fix there too

### DIFF
--- a/n_request.c
+++ b/n_request.c
@@ -161,6 +161,7 @@ bool NoteRequestWithRetry(J *req, uint32_t timeoutSeconds)
             // Free error response
             if (rsp != NULL) {
                 JDelete(rsp);
+                rsp = NULL;
             }
         } else {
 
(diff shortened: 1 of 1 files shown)


russelldeguzman:
n_request.c:163 in `NoteRequestWithRetry` has the same logic. Should we fix that one too while we're at it? It looks like we use that a few places in our code as well.

gwolffblues (approved):

gwolffblues:
Fix ensures responses from Notecard are set to NULL if the content of the response has already been deleted.

PR #18, opened 2022-11-02
Remove ASCII check to allow for UTF-8 characters

alvarop:
The `const char *serialNoteTransaction(char *json, char **jsonResponse)` function in n_serial.c discards data which contains any character which isn't ASCII (values 1-127). Unfortunately, this breaks if any field contains UTF-8 characters outside that range. Documentation says UTF-8 is supported, so I assume this is left over from before (https://dev.blues.io/hardware/notecard-datasheet/note-wbex-500/ )

Here's an example `{"req":"card.time"}` response which may cause the issue: `{"minutes":-180,"lat":-36.7327225,"lon":-73.15059765625,"area":"Talcahuano Biobío Region","country":"CL","zone":"-03,America/Santiago","time":1665850318}`

The í in Biobío is a UTF-8 character expressed with the following byte sequence: `0xc3 0xad`, which causes the error "serial communications error {io}"


alvarop:
You could do a more complex check to see if characters are valid UTF-8, but that would be a much bigger change.

alvarop:
Ran into the same issue with another unit with area "Valparaíso Valparaiso Region"

PR #21, opened 2022-11-30
Add initial unit testing.

haydenroche5:
This commit adds a couple very simple unit tests as a PoC. The tests use GoogleTest, which, to my knowledge, is the most mature and well-maintained open-source C++/C testing framework available. GTest supports mocking as well, which may come in handy when we start testing note-c functions that expect to interact with an actual Notecard. For now, the simple tests in this commit only exercise code that doesn't need a Notecard (e.g. allocating a creating a new Notecard request object).


haydenroche5:
Closing in favor of https://github.com/blues/note-c/pull/22.

PR #22, opened 2022-12-07
Add initial unit testing using Catch2 and fff.

haydenroche5:


zfields (approved):
Extraordinarily elegant solution for weak linking tested APIs!

m-mcgowan (commented):
Well-crafted solution!

PR #23, opened 2022-12-09
Initialize responseJSON to NULL in NoteTransaction.

haydenroche5:
Prior to this commit, if no response was expected in NoteTransaction, responseJSON would be a garbage value. If JParse then failed, we'd get a fault when we tried to free responseJSON. The solution is to initialize responseJSON to NULL.


zfields (approved):

zfields:
Good catch!

PR #25, opened 2022-12-11
Add GitHub action to run unit tests with every PR.

haydenroche5:
~~I opted not to Dockerize this, in contrast to what's been done in note-arduino. The build already runs in a GitHub-provisioned VM, so the addition of a container inside that seems like overkill.~~

Zak and I talked, and I decided to go the Docker/container route in the end, for a few reasons:

1) No hidden dependencies. For example, the GitHub VMs come with CMake installed, among other packages we need. With the Dockerfile, we fully enumerate our dependencies, so nothing is implied or hidden.
2) Easier for developers to run the GitHub action locally. This is mainly due to 1: the container will set up whatever dependencies it needs; we don't need to install anything up front.
3) Easier to migrate to a different CI platform if we ever outgrow GitHub actions.

One key downside is that, for the moment, we're building the Docker image anew on every CI run. When we eventually get set up with a container registry (e.g. [Github's container registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)), this problem will go away, as we'll be able to pull down images from the registry.

### DIFF
--- a/.github/actions/run_unit_tests/Dockerfile
+++ b/.github/actions/run_unit_tests/Dockerfile
@@ -1,0 +1,42 @@
+# Container with dependencies necessary to run note-c unit tests.
+
+# Build development environment
+# $ docker build . --tag note_c_run_unit_tests
+
+# Launch development environment (mount source root as /note-c/)
+# $ docker run --rm --volume $(pwd)/../../../:/note-c/ --workdir /note-c/ note_c_run_unit_tests
+
+# POSIX compatible (Linux/Unix) base image.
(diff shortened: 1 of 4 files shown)


m-mcgowan (approved):
I like the approach, simple and to the point.

m-mcgowan:
Not wanting to add more to the scope, but how about adding code coverage reporting to Coveralls? You can probably borrow from the existing implementation in `note-arduino`, recast as CMake scripts.

m-mcgowan (approved):
LGTM

PR #27, opened 2022-12-13
Add test coverage support with gcov and lcov.

haydenroche5:
I've not hooked this up to coveralls.io, yet. I want to get the coverage into a bit better shape before we start publishing the numbers. :smile:

### DIFF
--- a/.github/actions/run_unit_tests/Dockerfile
+++ b/.github/actions/run_unit_tests/Dockerfile
@@ -16,6 +16,7 @@ RUN ["dash", "-c", "\
      build-essential \
      ca-certificates \
      curl \
+     lcov \
  && apt-get clean \
  && apt-get purge \
  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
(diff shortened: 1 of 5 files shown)


zfields:
I have always used `gcovr` (only because I stumbled across it first). However, I understand that `lcov` is more commonly supported.

I am fully supportive of moving forward with `lcov`. That being said, do you think it's worth doing some kind of comparative analysis (like 15 minutes of research), or are you already familiar with both and `lcov` is superior?

haydenroche5:
> I have always used `gcovr` (only because I stumbled across it first). However, I understand that `lcov` is more commonly supported.
> 
> I am fully supportive of moving forward with `lcov`. That being said, do you think it's worth doing some kind of comparative analysis (like 15 minutes of research), or are you already familiar with both and `lcov` is superior?

I actually initially wrote this to use gcovr rather than lcov. I switched to lcov because note-arduino is using it in run_all_tests.sh. I'm happy to use either one. I think it just comes down to which HTML we like more, as they seem to produce largely the same data. [Here's an issue on the gcovr repo](https://github.com/gcovr/gcovr/issues/132) where one of the gcovr devs essentially says the same.

zfields (approved):
Side channel conversation

PR #30, opened 2022-12-20
Add a script to format all version-controlled source files with astyle. 

haydenroche5:
Also, add a GitHub action to ensure all commits use properly formatted code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,3 +14,12 @@ jobs:
         uses: actions/checkout@v3
       - name: Run unit tests
         uses: ./.github/actions/run_unit_tests
+  run_astyle:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
+      - name: Install astyle
+        run: sudo apt-get install -y astyle
+      - name: Check formatting
+        run: ./scripts/run_astyle.sh --check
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -14,3 +14,12 @@ jobs:
         uses: actions/checkout@v3
       - name: Run unit tests
         uses: ./.github/actions/run_unit_tests
+  run_astyle:
```
haydenroche5, line 17: I want to Dockerize this, too, eventually, just as we did for running the unit tests. However, we aren't set up with a container registry, yet, and I want this job to run and give feedback quickly. That's at odds with building the image anew on every run. So, for now, I'm not Dockerizing this one.

n_ftoa.c
```diff
@@ -287,8 +287,7 @@ fmtflt(char *str, size_t *len, size_t size, JNUMBER fvalue, int width,
         if (fpos > 0)	/* Omit trailing fractional part zeros. */
             while (omitcount < fpos && fconvert[omitcount] == '0') {
                 omitcount++;
-            }
-        else {	/* The fractional part is zero, omit it completely. */
+            } else {	/* The fractional part is zero, omit it completely. */
```
zfields, line 290: tabs / spaces

zfields (approved):

n_ftoa.c
zfields: which ironically seems like it should have been picked up by `astyle`... :raised_eyebrow:

haydenroche5: Fixed and added braces to the opening `if`, which I think makes the code clearer.

zfields:
Merge it already.

PR #31, opened 2022-12-22
Simplify NoteRequest and NoteRequestWithRetry.

haydenroche5:
These functions are nearly identical to NoteRequestResponse and NoteRequestResponseWithRetry, respectively. The only difference is that they delete the Notecard's response and return a bool. This commit modifies NoteRequest to use NoteRequestResponse internally. Same thing for the retry version.

### DIFF
--- a/n_request.c
+++ b/n_request.c
@@ -106,20 +106,15 @@ J *NoteNewCommand(const char *request)
 /**************************************************************************/
 bool NoteRequest(J *req)
 {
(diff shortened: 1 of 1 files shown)


zfields (commented):
I think this is a good change. Let's have Ray have a look at it to see if he agrees.

rayozzie:
Looks good.  Hope you don't screw it up :-)

btw Zak, did you ever get arduino Uno working well? Challenging.

PR #38, opened 2023-01-11
Add more unit tests and other cleanups.

haydenroche5:
- Fix some typos in n_cjson_helpers.c Doxygen comments.
- Check return value of _SerialReset in retry loop of serialNoteReset.
- Exclude third party code (e.g. n_md5.c) from coverage report.
- Remove unnecessary "inLog" code from NotePrint.
- Get rid of the need for some casting in i2cNoteReset and serialNoteTransaction.
- Add new unit tests for serialNoteTransaction, serialNoteReset, NoteSetFn, NoteDebug, NoteTransaction hooks, NoteSerial hooks, NotePrint, and NotePrintln.

### DIFF
--- a/n_serial.c
+++ b/n_serial.c
@@ -48,7 +48,7 @@ const char *serialNoteTransaction(char *json, char **jsonResponse)
         if (segLen > CARD_REQUEST_SERIAL_SEGMENT_MAX_LEN) {
             segLen = CARD_REQUEST_SERIAL_SEGMENT_MAX_LEN;
         }
-        _SerialTransmit((uint8_t *)&transmitBuf[segOff], segLen, false);
+        _SerialTransmit(&transmitBuf[segOff], segLen, false);
         segOff += segLen;
         segLeft -= segLen;
         if (segLeft == 0) {
@@ -212,8 +212,9 @@ bool serialNoteReset()
         _Debug("no notecard\n");
 #endif
         _DelayMs(500);
-        _SerialReset();
-
+        if (!_SerialReset()) {
+            return false;
+        }
     }
 
     // Done
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,14 @@ if(BUILD_TESTS)
          NoteI2CReceive;
          NoteLockI2C;
          NoteUnlockI2C;
-         NoteI2CReset"
+         NoteI2CReset;
+         NoteSerialAvailable;
+         NoteSerialTransmit;
+         NoteSerialReceive;
+         NoteSerialReset;
+         NoteIsDebugOutputActive;
+         NoteDebug;
+         NotePrint"
     )
     foreach(MOCKED_FN ${MOCKED_FNS})
         string(APPEND OBJCOPY_WEAKEN "-W ${MOCKED_FN} ")
--- a/n_cjson_helpers.c
+++ b/n_cjson_helpers.c
@@ -83,7 +83,7 @@ J *JGetObject(J *rsp, const char *field)
 
 //**************************************************************************/
 /*!
-    @brief  Return the boolean repsentation of an item.
+    @brief  Return the boolean representation of an item.
     @param   item The JSON item.
     @returns The boolean value.
 */
@@ -98,7 +98,7 @@ bool JBoolValue(J *item)
 
 //**************************************************************************/
 /*!
-    @brief  Return the string repsentation of an item.
+    @brief  Return the string representation of an item.
     @param   item The JSON item.
     @returns The string value, or empty string, if NULL.
 */
@@ -113,7 +113,7 @@ char *JStringValue(J *item)
 
 //**************************************************************************/
 /*!
-    @brief  Return the number repsentation of an item.
+    @brief  Return the number representation of an item.
     @param   item The JSON item.
     @returns The number, or 0.0, if NULL.
 */
@@ -151,7 +151,7 @@ JNUMBER JGetNumber(J *rsp, const char *field)
 
 //**************************************************************************/
 /*!
-    @brief  Return the integer repsentation of an item.
+    @brief  Return the integer representation of an item.
     @param   item The JSON item.
     @returns The number, or 0, if NULL.
 */
--- a/n_helpers.c
+++ b/n_helpers.c
@@ -182,19 +182,16 @@ void NotePrintln(const char *line)
 bool NotePrint(const char *text)
 {
     bool success = false;
+
     if (NoteIsDebugOutputActive()) {
         NoteDebug(text);
         return true;
     }
-    int inLog = 0;
-    if (inLog++ != 0) {
-        inLog--;
-        return false;
-    }
+
     J *req = NoteNewRequest("card.log");
     JAddStringToObject(req, "text", text);
     success = NoteRequest(req);
-    inLog--;
+
     return success;
 }
 
@@ -328,7 +325,7 @@ bool NoteRegion(char **retCountry, char **retArea, char **retZone, int *retZoneO
     if (retZoneOffset != NULL) {
         *retZoneOffset = curZoneOffsetMins;
     }
-    return true;;
+    return true;
 }
 
 //**************************************************************************/
--- a/n_i2c.c
+++ b/n_i2c.c
@@ -42,14 +42,14 @@ static void _DelayIO()
   @param   jsonResponse
   An out parameter c-string buffer that will contain the JSON
   response from the Notercard.
-  @returns a c-string with an error, or `NULL` if no error ocurred.
+  @returns a c-string with an error, or `NULL` if no error occurred.
 */
 /**************************************************************************/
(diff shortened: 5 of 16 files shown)


n_serial.c
```diff
@@ -48,7 +48,7 @@ const char *serialNoteTransaction(char *json, char **jsonResponse)
         if (segLen > CARD_REQUEST_SERIAL_SEGMENT_MAX_LEN) {
             segLen = CARD_REQUEST_SERIAL_SEGMENT_MAX_LEN;
         }
-        _SerialTransmit((uint8_t *)&transmitBuf[segOff], segLen, false);
+        _SerialTransmit(&transmitBuf[segOff], segLen, false);
```
zfields, line 51: Based on the type of `transmitBuf` above, this looks like a nice clean-up. :+1:

But just to confirm, the tests running with `-Wall`, `-Wextra`, `-Wpedantic`, `-Werror`, right?

The reason I ask, is because a lot of the casting you see throughout the code was to keep different compilers from issuing warnings about our code. This one seems to be a non-issue, but it made me think about us having our flags enabled.

_I didn't look too hard, but I couldn't immediately find them the `CMakeLists.txt` below._

zfields (changes requested):

rayozzie:
It’s yet another hack-y way of getting debug output that is probably undocumented and you can probably desupport if you like.

Where it has come in useful is this:

  *   Someone develops an app with note-c (or note-arduino, or whatnot)
  *   Because of the nature of their hardware, they have no “Arduino debug console” or other such thing where they can see debug output on their mcu.  (We had this problem badly with, I believe, some little msp432 we were playing with.)  BUT they are desperate to see what is happening inside their host MCU
  *   The NoteDebug, when hooks aren’t registered, will call card.debug and output the app’s debug output on the notecarrier’s “trace mode” usb console
  *   The reason for “inLog” is self-evident if you were to remove it:  NoteTransaction ITSELF does NoteDebug output for transaction debugging, and without the flag we’d go recursive and blow the stack

From: Zachary J. Fields ***@***.***>
Date: Thursday, January 12, 2023 at 6:13 AM
To: blues/note-c ***@***.***>
Cc: Ray Ozzie ***@***.***>, Mention ***@***.***>
Subject: Re: [blues/note-c] Add more unit tests and other cleanups. (PR #38)

@zfields requested changes on this pull request.

________________________________

In n_helpers.c<https://github.com/blues/note-c/pull/38#discussion_r1068105717>:

> @@ -186,15 +186,9 @@ bool NotePrint(const char *text)



         NoteDebug(text);



         return true;



     }



-    int inLog = 0;



@rayozzie<https://github.com/rayozzie> , Can you remind us of the original intent of the inLog variable?

Obviously, it wasn't doing anything as it stood, but is there some functionality we should have and are missing?

________________________________

In n_i2c.c<https://github.com/blues/note-c/pull/38#discussion_r1068124854>:

> @@ -242,14 +242,14 @@ bool i2cNoteReset()



     for (retries=0; transmitErr==NULL && !notecardReady && retries<3; retries++) {







         // Loop to drain all chunks of data that may be ready to transmit to us



-        int chunklen = 0;



+        uint32_t chunklen = 0;


⬇️ Suggested change

-        uint32_t chunklen = 0;



+        uint16_t chunklen = 0;



I'm always leary of changing types.

Moving to unsigned is a good change, and I cannot see any immediate side-effects of uint32_t. However, since you are making the change, perhaps you can select a type that is "guaranteed to be safe", like uint16_t.

The signature for _I2CReceive() is:

const char *NoteI2CReceive(uint16_t DevAddress, uint8_t* pBuffer, uint16_t Size, uint32_t *available)

chunklen goes into the Size parameter, which is uint16_t.

As note-c written today, the value in chunklen cannot exceed 127, so there is no problem. That being said, why not play is safe?

________________________________

In n_serial.c<https://github.com/blues/note-c/pull/38#discussion_r1068148716>:

> @@ -212,8 +212,9 @@ bool serialNoteReset()



         _Debug("no notecard\n");



 #endif



         _DelayMs(500);



-        _SerialReset();



Whoa. Does this change behavior?

I'm concerned, because note-c is used by all our customers. If this is a bug, then it probably needs to be documented and stay a bug, because it's likely either depended upon or already been worked around, and the worst thing we could do is break our customers code.

@rayozzie<https://github.com/rayozzie> Can you bless this change?

________________________________

In n_serial.c<https://github.com/blues/note-c/pull/38#discussion_r1068159534>:

> @@ -48,7 +48,7 @@ const char *serialNoteTransaction(char *json, char **jsonResponse)



         if (segLen > CARD_REQUEST_SERIAL_SEGMENT_MAX_LEN) {



             segLen = CARD_REQUEST_SERIAL_SEGMENT_MAX_LEN;



         }



-        _SerialTransmit((uint8_t *)&transmitBuf[segOff], segLen, false);



+        _SerialTransmit(&transmitBuf[segOff], segLen, false);



Based on the type of transmitBuf above, this looks like a nice clean-up. 👍

But just to confirm, the tests running with -Wall, -Wextra, -Wpedantic, -Werror, right?

The reason I ask, is because a lot of the casting you see throughout the code was to keep different compilers from issuing warnings about our code. This one seems to be a non-issue, but it made me think about us having our flags enabled.

I didn't look too hard, but I couldn't immediately find them the CMakeLists.txt below.

________________________________

In n_i2c.c<https://github.com/blues/note-c/pull/38#discussion_r1068144640>:

> @@ -242,14 +242,14 @@ bool i2cNoteReset()



     for (retries=0; transmitErr==NULL && !notecardReady && retries<3; retries++) {







         // Loop to drain all chunks of data that may be ready to transmit to us



-        int chunklen = 0;



+        uint32_t chunklen = 0;



It's odd that the available parameter returns a uint32_t, which is then loaded into chunklen. 🤔

NoteI2cReceive is executing a callback (beyond our control), so it seems like we should program defensively and also add a bounds check (along with test) around this edge case.

Below (line #266) you could change the logic to protect against it with:

            // Read everything that's left on the module



            chunklen = (available > USHRT_MAX) ? USHRT_MAX : available;

n_serial.c
haydenroche5: It's in the top-level CMakeLists.txt:

```
add_compile_options(
    -Wall
    -Wextra
    -Wpedantic
    -Werror
)
```

n_serial.c
```diff
@@ -212,8 +212,9 @@ bool serialNoteReset()
         _Debug("no notecard\n");
 #endif
         _DelayMs(500);
-        _SerialReset();
```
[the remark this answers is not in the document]
haydenroche5: A couple years ago Ray changed the serial and I2C reset functions to return a bool:

```
commit 808dd1edca52e28b03fda10af0d59bd3dfccafb8
Author: Ray Ozzie <ray@ozzie.net>
Date:   Wed Aug 12 10:02:21 2020 -0400

    fix

diff --git a/n_lib.h b/n_lib.h
index a8cc84f..ac90f15 100644
--- a/n_lib.h
+++ b/n_lib.h
@@ -40,11 +40,11 @@ bool serialNoteReset(void);
 // Hooks
 void NoteLockNote(void);
 void NoteUnlockNote(void);
-void NoteSerialReset(void);
+bool NoteSerialReset(void);
 void NoteSerialTransmit(uint8_t *, size_t, bool);
 bool NoteSerialAvailable(void);
 char NoteSerialReceive(void);
-void NoteI2CReset(void);
+bool NoteI2CReset(void);
 const char *NoteI2CTransmit(uint16_t DevAddress, uint8_t* pBuffer, uint16_t Size);
 const char *NoteI2CReceive(uint16_t DevAddress, uint8_t* pBuffer, uint16_t Size, uint32_t *avail);
 bool NoteHardReset(void);
```

It seems to me that we wouldn't want to do the retry loop again if we're not able to reset the serial bus. Indeed, that's what's already happening at the start of `serialNoteReset`:

```
if (!_SerialReset()) {
    return false;
}
```

> because it's likely either depended upon or already been worked around

I think it's likely neither of those, actually. As I said, if we've made it this far `_SerialReset` already succeeded once. It seems unlikely to me that it would fail again here, so I would guess no one's actually hitting this in "real life." Also, if it did fail, I would think the bus is in an unusable state, so retrying would have no effect anyway, as transmit/receive aren't going to work.

n_serial.c
zfields: I thought I remembered you doing that, but I couldn't find it. :+1:

haydenroche5:
> The reason for “inLog” is self-evident if you were to remove it:  NoteTransaction ITSELF does NoteDebug output for transaction debugging, and without the flag we’d go recursive and blow the stack

Correct, but the function in question is `NotePrint`, not `NoteDebug`. `NoteTransaction` should never call `NotePrint`, from my reading of the code.

rayozzie:
get rid of the inLog then.  it must be historical before NoteDebug was separated from NotePrint

zfields (approved):

PR #40, opened 2023-01-20
Feature/fix getms overflow

MarkoSagadin:
Greetings!

At IRNAS we are building a product with Notecards and we have decided to build our Zephyr driver on top of you note-c library. 

During code review of my Notecard driver code one of my colleague commented that I should check, if the note-c library correctly handles overflow that would happen every ~49 days in below code:

```c
/* zephyr_millis is passed to note-c as a implementation of NoteGetMs() */
static uint32_t zephyr_millis(void)
{
	return (uint32_t)k_uptime_get();
}

static int notecard_init(const struct device *dev)
{
	k_heap_init(&notecard_heap, notecard_heap_memory, CONFIG_NOTECARD_HEAP_SIZE);
	k_mutex_init(&notecard_mutex);

	NoteSetFnDebugOutput(zephyr_log_print);

	/* Set platform specific hooks. */
	NoteSetFn(zephyr_malloc, zephyr_free, zephyr_delay, zephyr_millis);
	return 0;
}
```

I checked the note-c code and I saw that this is not correctly handled in (almost) all uses of `NoteGetMs()` (defined as `_GetMs()`).

When using timing functions that can rollover to measure elapsed time, the difference approach should be used; that way overflows are ignored and elapsed duration is correctly calculated.

I went ahead and fixed the uses of _GetMs(). I used `scripts/run_unit_tests.sh` to run the unit tests.

I made several commit pairs in the following convention: 
* first commit in a pair introduces a new unit test that trigger overflow condition, because of it the unit tests hangs.
* second commit fixes the overflow.

Three things to note:
* I am not sure that I covered all possible execution paths in the tests, I learned about one hour ago how Catch2 `SECTION` macro works and I am sure that added tests could be written differently to cover more possible paths due to other `SECTION`s.
* I did not change logic in `NoteTimeST()` which uses `_GetMs()`. It looks like it handles the wraparound, but I would suggest that someone looks over it again.
* I ran the the `scripts/run_astyle.sh` before commiting, but it did nothing, I am not sure if it works or maybe I followed your style in the code that well that no changes are needed.

### DIFF
--- a/n_i2c.c
+++ b/n_i2c.c
@@ -191,7 +191,7 @@ const char *i2cNoteTransaction(char *json, char **jsonResponse)
         }
 
         // If we've timed out and nothing's available, exit
-        if (_GetMs() >= startMs + (NOTECARD_TRANSACTION_TIMEOUT_SEC*1000)) {
+        if (_GetMs() - startMs >= NOTECARD_TRANSACTION_TIMEOUT_SEC*1000) {
             _Free(jsonbuf);
 #ifdef ERRDBG
             _Debug("reply to request didn't arrive from module in time\n");
--- a/n_request.c
+++ b/n_request.c
@@ -203,7 +203,8 @@ J *NoteRequestResponseWithRetry(J *req, uint32_t timeoutSeconds)
     J *rsp;
 
     // Calculate expiry time in milliseconds
-    uint32_t expiresMs = _GetMs() + (timeoutSeconds * 1000);
+    uint32_t expiresMs = _GetMs();
+    uint32_t timeouMilliSeconds = timeoutSeconds * 1000;
 
     while(true) {
         // Execute the transaction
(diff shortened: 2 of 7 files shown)


haydenroche5:
Thanks for the fixes @MarkoSagadin! I made some changes, mostly to the testing code, and since I can't add commits to your branch, I opened a new PR with all your commits here + my changes: https://github.com/blues/note-c/pull/46.

MarkoSagadin:
Thanks for merging this! If I open some other PR in the future, should I change something in the process to make the review, merging, etc. smoother?

haydenroche5:
The process we used here is ok with me. :)

haydenroche5:
@MarkoSagadin Since you mentioned you're using Zephyr, I thought you might find this useful: https://github.com/blues/note-zephyr

We've been improving our Zephyr integration lately. Cheers. :)

MarkoSagadin:
@haydenroche5 Thanks, this was a starting point for my driver :)

PR #42, opened 2023-01-25
Add more unit tests.

haydenroche5:
Adds tests for:

- NoteTime
- NoteRegion
- NoteLocationValid
- NoteIsConnected
- NoteGetNetStatus
- NoteGetVersion
- NoteGetLocation
- NoteSetLocation
- NoteClearLocation
- NoteGetLocationMode
- NoteSetLocationMode
- NoteGetServiceConfig
- NoteGetStatus
- NoteSleep


n_helpers.c
```diff
@@ -754,7 +754,7 @@ bool NoteGetLocation(JNUMBER *retLat, JNUMBER *retLon, JTIME *time, char *status
     J *rsp = NoteRequestResponse(NoteNewRequest("card.location"));
     if (rsp != NULL) {
         if (statusBuf != NULL) {
-            strlcpy(statusBuf, JGetString(rsp, "err"), statusBufLen);
+            strlcpy(statusBuf, JGetString(rsp, "status"), statusBufLen);
```
zfields, line 757: This looks like a change in behavior.

Can you create an example of what our customers would have been seeing, and also show what the new behavior will produce?

Even when public APIs behave in a way that is "wrong", there may be a customer algorithm depending on that "wrong" behavior. This would be a very difficult bug to discover, if your application had previously worked for the last several versions of `note-c`.

We should probably isolate this change into it's own PR, and seek input from Ray.

zfields (changes requested):

n_helpers.c
haydenroche5: I think this may have just a typo. If you look in req.c in the Notecard firmware, there's no "err" field that gets added to the response to `card.location`, but there is a status field.

> Can you create an example of what our customers would have been seeing, and also show what the new behavior will produce?

`JGetString(rsp, "err")` will return the empty string, because that field doesn't exist, and that's what they'll get back in their passed in `statusBuf`.

> We should probably isolate this change into it's own PR, and seek input from Ray.

Sure.

haydenroche5: Separate PR up here: https://github.com/blues/note-c/pull/43.

zfields (approved):

PR #45, opened 2023-02-01
Add optional memory checking with valgrind.

haydenroche5:
Turn it on for the GitHub action that runs unit tests. Also, resolve a memory leak in NoteGetEnv_test.cpp.

### DIFF
--- a/scripts/run_unit_tests.sh
+++ b/scripts/run_unit_tests.sh
@@ -1,10 +1,12 @@
 #!/bin/bash
 
 COVERAGE=0
+MEM_CHECK=0
 
 while [[ "$#" -gt 0 ]]; do
     case $1 in
         --coverage) COVERAGE=1 ;;
+        --mem-check) MEM_CHECK=1 ;;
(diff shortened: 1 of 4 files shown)


scripts/run_unit_tests.sh
```diff
@@ -34,14 +41,14 @@ if [[ $? -ne 0 ]]; then
     exit 1
 fi
 
-cmake --build build/ -- $BUILD_OPTIONS
+cmake --build build/ -- $BUILD_OPTIONS -j
```
zfields, line 44: :raised_hands:

test/src/NoteGetEnv_test.cpp
```diff
@@ -74,6 +76,8 @@ TEST_CASE("NoteGetEnv")
         CHECK(NoteRequestResponse_fake.call_count == 1);
         // Ensure the default value was copied into outBuf.
         CHECK(!strcmp(outBuf, defaultVal));
+
+        JDelete(req);
```
zfields, line 80: :joy:

zfields (approved):

PR #53, opened 2023-02-28
Clarify NoteRequest return when sending 'cmd'

sdt99:
Clarify in function comments that a `true` return value from NoteRequest when sending a `cmd` indicates that the `cmd` was sent without error, but does not guarantee that the Notecard received and processed the `cmd`.
The application should make other provisions to determine that the `cmd` was processed if this is critical to system behavior.


sdt99:
One thought that occurs to me is whether NoteRequest() should check for an error response if sending a `cmd`:

```
{"cmd":"hub.get"}
{"cmd":"hub.g
{"err":"cannot interpret JSON: premature end of input near '\"hub.g ' {io}"}
```

This would add value in a situation where a critical cmd (such as a card.attn with a payload) is sent - but it's probably low priority since when the host sends a critical `cmd` it should verify it was processed and retry if needed.

PR #54, opened 2023-03-04
Add unit tests to cover n_ua.c.

haydenroche5:
Also, fix a couple problems in n_ua.c:

- The logic for putting the Atollic GCC compiler in the user agent object was written such that it would never trigger.
- We allow the user to set n_cpu_name via NoteSetUserAgentCPU, but then in NoteUserAgent we were overriding whatever they set it to, based on preprocessor logic. I moved that preprocessor logic out so that it controls the default value of n_cpu_name, but if the user overrides it, that's what we put in the user agent object.


n_ua.c
```diff
@@ -25,7 +25,49 @@ static int n_cpu_mem = 0;
 static int n_cpu_mhz = 0;
 static int n_cpu_cores = 0;
 static char *n_cpu_vendor = NULL;
+
 static char *n_cpu_name = NULL;
+#if defined(ARDUINO_ARCH_ARC32)
```
zfields, line 30: Why change the whitespace? Was this done by the linter?

n_ua.c
```diff
@@ -75,10 +117,10 @@ __attribute__((weak)) J *NoteUserAgent()
     char *compiler = (char *) ("iar" PLUS " " STRINGIFY(__VER__));
 #elif defined(__clang__)
     char *compiler = (char *) ("clang" PLUS " " __VERSION__);
-#elif defined(__GNUC__)
-    char *compiler = (char *) ("gcc" PLUS " " __VERSION__);
 #elif defined(__ATOLLIC__) && defined(__GNUC__)
     char *compiler = (char *) ("atollic gcc" PLUS " " __VERSION__);
+#elif defined(__GNUC__)
```
zfields, line 122: This change is fine, but I will take this opportunity to belly ache that I would like to see modifications performed on the underlying library as separate PRs from unit-tests.

To me it seems cleaner, and maintains the separation of church and state.

zfields (approved):

n_ua.c
haydenroche5: I'm fine with that going forward.

n_ua.c
haydenroche5: Yes. astyle removed the indentation.

PR #62, opened 2023-03-31
fix: assignment error

zfields:
```none
warning: data definition has no type or storage class
warning: type defaults to 'int' in declaration of 'n_cpu_name' [-Wimplicit-int]
error: conflicting types for 'n_cpu_name'; have 'int'
```

Variables may only be declared at global scope, not assigned.

Fix adds type prefix to change from assignment to declaration.


zfields:
Validated the fix works: https://github.com/blues/note-arduino/actions/runs/4575067278

haydenroche5 (approved):

PR #63, opened 2023-04-07
Add a wrapper around NoteDebug that supports log level and __FILE__/__LINE__.

haydenroche5:


note.h
```diff
@@ -169,6 +169,33 @@ void NoteDebug(const char *message);
...
+    __FILE__, __LINE__, __VA_ARGS__)
+#define NOTE_C_LOG_DEBUG(...) NoteDebugWithLevel(NOTE_C_LOG_LEVEL_DEBUG, \
+    __FILE__, __LINE__, __VA_ARGS__)
+
+#ifndef NOTE_C_DEBUG_W_LEVEL_DEFAULT_MAX
+#define NOTE_C_DEBUG_W_LEVEL_DEFAULT_MAX NOTE_C_LOG_LEVEL_ERROR
+#endif
+
+#ifndef NOTE_C_DEBUG_W_LEVEL_MAX_BYTES
+#define NOTE_C_DEBUG_W_LEVEL_MAX_BYTES 256
```
zfields, line 196: 256 may be too big.

We are already suffering from space constraints with `note-c`. There is a precedent set by another debugging buffer. I think it's 128 or 64. We should use the same `#define` it uses for it's size, or declare one they both can share.

test/src/NoteDebug_test.cpp
```diff
@@ -21,7 +21,7 @@ namespace
 {
 
 typedef struct {
-    char debugBuf[32];
+    char debugBuf[256];
```
zfields, line 24: Is it possible to `#include "note.h"` here, or will it go haywire?

```suggestion
    char debugBuf[NOTE_C_DEBUG_W_LEVEL_MAX_BYTES];
```

n_hooks.c
```diff
@@ -418,6 +418,95 @@ void NoteDebug(const char *line)
 #endif
 }
 
+static uint8_t maxLogLevel = NOTE_C_DEBUG_W_LEVEL_DEFAULT_MAX;
+
+//**************************************************************************/
+/*!
+  @brief        Write to the debug stream, prepending the message with the file
+                and line number where this function was called. If the level is
+                below the maxLogLevel, the message will be dropped.
```
zfields, line 427:
```suggestion
                above the maxLogLevel, the message will be dropped.
```

note.h
zfields: No, it looks like there are several examples of 256 byte buffers, but I still think this is excessive. Separately, we need to devise a scheme for managing all these buffers.

zfields (approved):

note.h
haydenroche5: Discussed over Slack.

PR #64, opened 2023-04-07
Namespace test macros with NOTE_C_ prefix.

haydenroche5:
Checked that the same number of tests were run before and after this change, to make sure I didn't miss any in the changeover to NOTE_C_TEST.

### DIFF
--- a/note.h
+++ b/note.h
@@ -73,15 +73,6 @@
 // originates from the Notecard, which synchronizes the time from both the cell network and GPS.
 typedef unsigned long int JTIME;
 
-// If we're building the tests, NOTE_C_STATIC is defined to nothing. This allows
-// the tests to access the static functions in note-c. Among other things, this
-// let's us mock these normally static functions.
-#ifdef NOTE_C_TEST
-#define NOTE_C_STATIC
-#else
-#define NOTE_C_STATIC static
-#endif
-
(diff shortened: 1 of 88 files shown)


note.h
```diff
@@ -73,13 +73,13 @@
...
+// If we're building the tests, NOTE_C_STATIC is defined to nothing. This allows
+// the tests to access the static functions in note-c. Among other things, this
 // let's us mock these normally static functions.
-#ifdef TEST
-#define STATIC
+#ifdef NOTE_C_TEST
+#define NOTE_C_STATIC
 #else
-#define STATIC static
+#define NOTE_C_STATIC static
```
zfields, line 82: Can this be moved into `n_helpers.c`, or into a test only header file that lives in `test/include`?

zfields (changes requested):

note.h
haydenroche5: > Can this be moved into n_helpers.c

I don't love the idea of moving it into n_helpers.c because other source files might have functions we want to mock at some point.

> or into a test only header file that lives in test/include

And then include that header in n_helpers.c (and elsewhere, if needed)? That could work. Please confirm my understanding, though.

zfields (approved):

PR #72, opened 2023-06-07
feat: JGetArray()

zfields:


zfields:
I ran into an issue where we don't have a helper to get arrays and I was mistakenly using `JGetObject()`, which resulted in a logic error.

@rayozzie I added you, so you could tell me if there is an alternative approach that I'm completely missing.
@haydenroche5 I added you, so you could tell me that I forgot to add test and you could point me at the test for `JGetObject()`

rayozzie:
This is fine. I am lazy and I probably would have just added
#define JGetArray JGetObjectItem
...but yours is fine as it does the type check.

zfields:
[Add `JGetArray()` to `note-c`](https://trello.com/c/0r9xZlbG/131-add-jgetarray-to-note-c)

haydenroche5:
> @haydenroche5 I added you, so you could tell me that I forgot to add test and you could point me at the test for JGetObject()

JGetObject_test.cpp

haydenroche5 (changes requested):
The test is going to be a near copy-paste of JGetObject_test.cpp.

haydenroche5 (approved):

PR #74, opened 2023-06-11
feat: add automatic crc support, supported by notecard firmware as of 2023-06-10

rayozzie:
The feature, which works only for 'requests' and not 'commands', works like this:
- every request's JSON is tagged with an additional field "crc" containing a sequence number and crc32 of the request
- notecard firmware prior to the new firmware ignores the extra json field
- the new firmware, as it sees this field, verifies the crc and sends an {io} error response back to the host if it doesn't match
- if the crc is present AND it matches, the firmware then attaches an analogous crc to the response back to note-c, containing the exact same sequence number that it received on the request
- note-c, when it receives the response, verifies this crc
- in note-c, if the sequence number is ever unexpected or the crc mismatches or it gets an {io} error, it automatically retries up to 10 times

Net-net, this should increase the robustness of host/notecard requests with no real downside.

### DIFF
--- a/note.h
+++ b/note.h
@@ -136,17 +136,32 @@ void NoteSetFnNoteMutex(mutexFn lockNotefn, mutexFn unlockNotefn);
 void NoteSetFnDefault(mallocFn mallocfn, freeFn freefn, delayMsFn delayfn, getMsFn millisfn);
 void NoteSetFn(mallocFn mallocfn, freeFn freefn, delayMsFn delayfn, getMsFn millisfn);
 void NoteSetFnSerial(serialResetFn resetfn, serialTransmitFn writefn, serialAvailableFn availfn, serialReceiveFn readfn);
-
-#define NOTE_I2C_ADDR_DEFAULT	0x17
-#ifndef NOTE_I2C_MAX_DEFAULT
-#define NOTE_I2C_MAX_DEFAULT	30
-#endif
-#ifndef NOTE_I2C_MAX_MAX
-#define NOTE_I2C_MAX_MAX		127
-#endif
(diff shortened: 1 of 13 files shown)


zfields:
I've updated the unit-tests to pass, but I need to review with @haydenroche5 to ensure this is a desired change.

note.h
```diff
@@ -136,17 +136,32 @@ void NoteSetFnNoteMutex(mutexFn lockNotefn, mutexFn unlockNotefn);
...
+// The Notecard, whose default I2C address is below, uses a serial-to-i2c protocol whose "byte count"
+// must fit into a single byte and which must not include a 2-byte header field.  This is why the
+// maximum that can be transmitted by note-c in a single I2C I/O is 255 - 2 = 253 bytes.
 #define NOTE_I2C_ADDR_DEFAULT	0x17
-#ifndef NOTE_I2C_MAX_DEFAULT
-#define NOTE_I2C_MAX_DEFAULT	30
-#endif
 #ifndef NOTE_I2C_MAX_MAX
-#define NOTE_I2C_MAX_MAX		127
+#define NOTE_I2C_MAX_MAX		253
```
zfields, line 148: Can we break `NOTE_I2C_MAX_MAX` into two parts?

```c
#define NOTE_I2C_HEADER_SIZE 2
#define NOTE_I2C_MAX_MAX (UCHAR_MAX - NOTE_I2C_HEADER_SIZE)
```

rayozzie: Yea that's fine go for it

haydenroche5:
@rayozzie When you have time, it would be good to get your review on my latest changes to `NoteTransaction`.

PR #80, opened 2023-06-15
Improve Doxygen comments in n_request.c.

haydenroche5:
Also, make them all conform to the same style.

This is a step toward generating the API docs for note-c. I'll tackle the other source files next, and eventually will come up with a scheme for generating and publishing the docs somewhere.

### DIFF
--- a/n_request.c
+++ b/n_request.c
@@ -1,14 +1,13 @@
 /*!
- * @file n_request.c
- *
- * Written by Ray Ozzie and Blues Inc. team.
- *
- * Copyright (c) 2019 Blues Inc. MIT License. Use of this source code is
- * governed by licenses granted by the copyright holder including that found in
- * the
- * <a href="https://github.com/blues/note-c/blob/master/LICENSE">LICENSE</a>
- * file.
- *
+ @file n_request.c
+
+ Written by Ray Ozzie and Blues Inc. team.
+
+ Copyright (c) 2019 Blues Inc. MIT License. Use of this source code is
+ governed by licenses granted by the copyright holder including that found in
+ the
+ <a href="https://github.com/blues/note-c/blob/master/LICENSE">LICENSE</a>
+ file.
  */
 
 #include "n_lib.h"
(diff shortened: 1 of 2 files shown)


n_request.c
```diff
@@ -587,20 +590,19 @@ void NoteErrorClean(char *begin)
...
+ * @param p The hex string to convert.
+ * @param len The length of the hex string.
+ *
+ * @returns The converted number.
+ */
+uint64_t n_atoh(char *p, int len)
 {
     uint64_t n = 0;
-    char *ep = p+maxlen;
+    char *ep = p+len;
```
haydenroche5, line 605: Not sure why this was called maxlen before. It's just the length of the passed in buffer, as far as I can tell. Max seems superfluous.

n_request.c
```diff
@@ -750,4 +751,5 @@ NOTE_C_STATIC bool crcError(char *json, uint16_t shouldBeSeqno)
     uint32_t shouldBeCrc32 = crc32(json, fieldOffset);
     return (shouldBeSeqno != actualSeqno || shouldBeCrc32 != actualCrc32);
 }
-#endif
+
+#endif // !NOTE_LOWMEM
```
haydenroche5, line 755: There was an `#ifdef` guard on each of the CRC functions. I simplified it by just using one for all of them as a group.

zfields:
Can you resolve the conflicts now that we've reworked a bunch of stuff?

n_request.c
zfields: I would guess it has to do with lines numbers 612, 613, and 614.

haydenroche5:
> Can you resolve the conflicts now that we've reworked a bunch of stuff?

Yep. Should be good to go now.

n_request.c
```diff
@@ -648,20 +644,19 @@ void NoteErrorClean(char *begin)
...
-/**************************************************************************/
-uint64_t n_atoh(char *p, int maxlen)
+ @brief Convert a hex string to a 64-bit unsigned integer.
+
+ @param p The hex string to convert.
+ @param maxLen The length of the hex string.
+
+ @returns The converted number.
+ */
+uint64_t n_atoh(char *p, int maxLen)
```
zfields, line 656: Did you also update the declaration?

zfields (approved):
Approved with comments.

n_request.c
```diff
@@ -570,63 +573,56 @@ J *noteTransactionShouldLock(J *req, bool lockNotecard)
...
-  @param   errstr
-  The error string.
-  @param   errtype
-  The error type string.
-  @returns boolean. `true` if the string contains the error provided, `false`
-  if not.
-*/
-/**************************************************************************/
+ @brief Check to see if a Notecard error is present in a JSON string.
+
```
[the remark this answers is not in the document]
zfields: I was hoping you would fashion that into the long description, similar to `NoteErrorClean()` below.

FWIW, I normally throw @notes at the bottom, not that it matters.

zfields (approved):

PR #85, opened 2023-07-12
feat: card.binary support

zfields:
//TODO:
* ~I still need to make better error messages surrounding "feature not implemented"~
* I have consolidated the logic around the receiving into a dynamic buffer, maybe should implement `_DynamicBufferReceive()`; or something like that.
* ~I need to make the User API to calculate the size of the total binary buffer for any given length of data.~
* I need to make Unit Tests for all the new functions and functionality.


n_cobs.c
```diff
@@ -0,0 +1,135 @@
...
+
+//**************************************************************************/
+/*!
+  @brief  Compute the encoding length of unencoded data
+
+  @param  ptr Pointer to the data to encode
+  @param  length Length of the data to encode
+
+  @return the length required for encoded data
+  @note  The computed length includes the NULL terminator
```
haydenroche5, line 97: Is it a NULL terminator or the EOP byte? If the latter, I think we should change the language to reflect that.

n_cobs.c
```diff
@@ -0,0 +1,135 @@
...
+//**************************************************************************/
+/*!
+  @brief Compute the maximum length of unencoded data that can fit into a
+  buffer of specified length.
+
+  @param  bufLen Length of the buffer in bytes
+
+  @return the length of unencoded data
+  @note The computation may leave additional space at the end, including one
+        byte for a NULL terminator.
```
haydenroche5, line 128: Same question re: NULL terminator.

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
 static const char *daynames[] = {"Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"};
 
 // Forwards
+NOTE_C_STATIC void setTime(JTIME seconds);
 NOTE_C_STATIC bool timerExpiredSecs(uint32_t *timer, uint32_t periodSecs);
 NOTE_C_STATIC int ytodays(int year);
 
+//**************************************************************************/
+/*!
+  @brief  Receive a large binary object from the Notecard's binary buffer
```
haydenroche5, line 88: I would remove "large", since it's not necessarily large.

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
...
+  @param  buffer A buffer to hold the result
```
haydenroche5, line 89: the result -> the binary object

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
...
+  @param  bufLen The total length of the buffer
+  @returns  NULL on success, else an error string pointer.
+  @note  Buffers are decoded in place, the original contents of the buffer will
```
haydenroche5, line 92: in place, the -> in place. The

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
...
+    J *rsp = NoteRequestResponse(NoteNewRequest("card.binary"));
+    if (!rsp) {
+        return ERRSTR("unable to issue binary request", c_err);
+    }
+
+    // Ensure the transaction doesn't return an error
+    // to confirm the binary feature is available
+    if (NoteResponseError(rsp)) {
+        JDelete(rsp);
+        return ERRSTR("feature not implemented", c_bad);
```
haydenroche5, line 108: I would rather get the actual error in the response as opposed to "feature not implemented". The feature might be implemented and the actual error could be something else. Not sure if there's an easy way to do that here, though, since the return type is a const string... maybe just log it with `_Debug`?

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
...
+    }
+
+    // Examine "cobs" from the response to evaluate the space
+    // required to hold the COBS encoded data on the Notecard.
+    const size_t cobs = JGetInt(rsp,"cobs");
+    JDelete(rsp);
+    if (!cobs) {
+        return ERRSTR("no data on notecard", c_err);
+    }
+    if (cobs > bufLen) {
```
haydenroche5, line 118: I know we already discussed this, but it still rubs me the wrong way that the caller would have to either

1. Allocate a buffer that they assume will be large enough.
2. Do `card.binary` manually themselves, get the length of buffer needed, allocate that, and finally call this function.

One solution I've seen in other open-source C libraries:

Have this function take a  `uint8_t **` instead of a `uint8_t *`. `buffer == NULL` is an error. `buffer != NULL && *buffer == NULL` means allocate the buffer for me and return it in `*buffer`. `buffer != NULL && *buffer != NULL` means I, the caller, have allocated the buffer already (heap, stack -- doesn't matter), so use that. `size_t bufLen` becomes `size_t * bufLen`. In the case this function allocates the buffer for the caller, we return the length in `*bufLen`. In the case the caller is providing the buffer, we read the length of said buffer from `*bufLen`.

If that seems too complicated, that's fair. There's a middle ground somewhere, I'm sure. But I do think there's some unnecessary friction here.

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
...
+        J *rsp = NoteRequestResponse(req);
+        if NoteResponseError(rsp) {
+            JDelete(rsp);
+            return ERRSTR("failed to initialize binary transaction", c_err);
+        }
+
+        // Examine "status" from the response to evaluate the MD5 checksum.
+        strlcpy(status, JGetString(rsp,"status"), NOTE_MD5_HASH_STRING_SIZE);
+        JDelete(rsp);
+    }
```
haydenroche5, line 141: Should we error out if we failed to allocate req?

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
...
+
+    // Issue `card.binary.get` and capture `"status"` from response
+    char status[NOTE_MD5_HASH_STRING_SIZE] = {0};
+    J *req = NoteNewRequest("card.binary.get");
+    if (req) {
+        JAddIntToObject(req, "cobs", cobs);
+
+        // Ensure the transaction doesn't return an error.
+        J *rsp = NoteRequestResponse(req);
+        if NoteResponseError(rsp) {
```
haydenroche5, line 133: `if NoteResponseError(rsp)` -> `if (NoteResponseError(rsp))`. I get that the way the macro is written this works, but I think it looks less weird with my suggested change.

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
...
+            JDelete(rsp);
+            return ERRSTR("failed to initialize binary transaction", c_err);
+        }
+
+        // Examine "status" from the response to evaluate the MD5 checksum.
+        strlcpy(status, JGetString(rsp,"status"), NOTE_MD5_HASH_STRING_SIZE);
+        JDelete(rsp);
+    }
+
+    // `NoteReceiveBytes()` of the COBS encoded bytes either in a single I/O of
```
haydenroche5, line 143: Idk what this comment means.

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
...
+    char status[NOTE_MD5_HASH_STRING_SIZE] = {0};
+    J *req = NoteNewRequest("card.binary.get");
+    if (req) {
+        JAddIntToObject(req, "cobs", cobs);
+
+        // Ensure the transaction doesn't return an error.
+        J *rsp = NoteRequestResponse(req);
+        if NoteResponseError(rsp) {
+            JDelete(rsp);
+            return ERRSTR("failed to initialize binary transaction", c_err);
```
haydenroche5, line 135: Need to unlock here.

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
...
+        return ERRSTR(errstr, c_err);
+    }
+
+    // Check buffer overflow condition
+    if (overflow) {
+        return ERRSTR("unexpected data available", c_err);
+    }
+
+    // Decode it in-place which is safe because decoding shrinks
+    const size_t decLen = cobsDecode(buffer,bufLen,'\n',buffer);
```
haydenroche5, line 163: Add spaces after the commas.

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
...
+    buffer[decLen] = '\0';
```
haydenroche5, line 164: Why?

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
...
+
+    // Verify MD5
+    char hash_string[NOTE_MD5_HASH_STRING_SIZE] = {0};
+    NoteMD5HashString(buffer, decLen, hash_string, NOTE_MD5_HASH_STRING_SIZE);
+    if (strncmp(hash_string, status, NOTE_MD5_HASH_STRING_SIZE)) {
+        return ERRSTR("data transfer error", c_err);
```
haydenroche5, line 170: I would change the error message to say the computed MD5 doesn't match the received MD5. No reason to not be specific about the failure, I think.

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
...
+    if (overflow) {
+        return ERRSTR("unexpected data available", c_err);
+    }
+
+    // Decode it in-place which is safe because decoding shrinks
+    const size_t decLen = cobsDecode(buffer,bufLen,'\n',buffer);
+    buffer[decLen] = '\0';
+
+    // Verify MD5
+    char hash_string[NOTE_MD5_HASH_STRING_SIZE] = {0};
```
haydenroche5, line 167: hash_string -> hashString

haydenroche5 (changes requested):

n_helpers.c
zfields: Okay, I'll will inspect the message to eliminate assumptions, as well as issue a `_Debug` message.

n_helpers.c
zfields: too much implementation detail, not enough high-level

n_helpers.c
zfields: This is arbitrary, but I like it. Maybe safety reasons. Will add comment.

n_hooks.c
```diff
@@ -821,16 +831,57 @@ bool NoteHardReset()
...
+
+/**************************************************************************/
+/*!
+  @brief  Receive bytes over from the Notecard using the currently-set
+  platform hook.
+  @param   buffer A buffer to receive bytes into.
+  @param   size (in/out)
+            - (in) The size of the buffer in bytes.
+            - (out) The length of the received data in bytes.
+  @param   delay Respect delay standard transmission delays.
```
haydenroche5, line 856: "Respect delay standard transmission delays." Needs rewording.

n_hooks.c
```diff
@@ -821,16 +831,57 @@ bool NoteHardReset()
...
+  @param   timeoutMs The maximum amount of time, in milliseconds, to wait for
+            serial data to arrive. Passing zero (0) disables the timeout.
```
haydenroche5, line 858: Remove serial since this can also use I2C as the underlying transport.

n_hooks.c
```diff
@@ -821,16 +831,57 @@ bool NoteHardReset()
...
+  @param   overflow (out) The contents did not fit inside the provided buffer.
```
haydenroche5, line 859: I would be more specific. If it's just what the comment here says it is, why isn't overflow a bool?

n_hooks.c
```diff
@@ -821,16 +831,57 @@ bool NoteHardReset()
...
+    return notecardRawReceive(buffer, size, delay, timeoutMs, overflow);
+}
+
+/**************************************************************************/
+/*!
+  @brief  Transmit bytes over to the Notecard using the currently-set
+  platform hook.
+  @param   buffer A buffer of bytes to transmit.
+  @param   size The count of bytes in the buffer to send
+  @param   delay Respect delay standard transmission delays.
```
haydenroche5, line 877: Needs rewording, as above.

n_hooks.c
```diff
@@ -821,16 +831,57 @@ bool NoteHardReset()
...
+  @param   timeoutMs The maximum amount of time, in milliseconds, to wait for
+            serial data to arrive. Passing zero (0) disables the timeout.
+  @param   overflow (out) The contents did not fit inside the provided buffer.
+  @returns  A c-string with an error, or `NULL` if no error ocurred.
+*/
+/**************************************************************************/
+const char *NoteRawReceive(uint8_t *buffer, size_t *size, bool delay, size_t timeoutMs, size_t *overflow)
+{
+    if (notecardRawReceive == NULL || hookActiveInterface == interfaceNone) {
```
haydenroche5, line 865: I don't quite understand why we need to check hookActiveInterface here and in NoteRawTransmit. Isn't it enough to check if the hook we're calling is NULL?

n_i2c.c
```diff
@@ -297,3 +207,161 @@ bool i2cNoteReset()
     // Done
     return notecardReady;
 }
+
+/**************************************************************************/
+/*!
+  @brief  Receive bytes over I2C to the Notecard.
```
haydenroche5, line 213: to -> from

n_i2c.c
```diff
@@ -297,3 +207,161 @@ bool i2cNoteReset()
...
+/**************************************************************************/
+/*!
+  @brief  Receive bytes over I2C to the Notecard.
+  @param   buffer
+            A buffer to receive bytes into.
+  @param   size (in/out)
+            - (in) The size of the buffer in bytes.
+            - (out) The length of the received data in bytes.
+  @param   delay
+            Respect delay standard transmission delays.
```
haydenroche5, line 220: Reword.

n_i2c.c
```diff
@@ -297,3 +207,161 @@ bool i2cNoteReset()
...
+  @param   timeoutMs
+            The maximum amount of time, in milliseconds, to wait for serial data
+            to arrive. Passing zero (0) disables the timeout.
```
haydenroche5, line 223: Ah, right, zero is 0. Thanks for the tip. :laughing:

n_i2c.c
```diff
@@ -297,3 +207,161 @@ bool i2cNoteReset()
...
+  @param   overflow (out)
+            The contents did not fit inside the provided buffer.
```
haydenroche5, line 225: Same comment as before. It's more than just a bool, at least from what I can tell.

n_i2c.c
```diff
@@ -297,3 +207,161 @@ bool i2cNoteReset()
...
+
+        // Look for end-of-packet marker
+        if (received > 0 && !eop) {
+            eop = (buffer[received-1] == '\n');
+        }
+
+        // Check overflow condition
+        *overflow = ((received + requested) >= *size);
+        if (*overflow) {
+            *overflow = available; // `available` should optimize allocation
```
haydenroche5, line 274: What's meant by the comment here?

n_i2c.c
```diff
@@ -297,3 +207,161 @@ bool i2cNoteReset()
...
+            // *overflow = requested;
```
haydenroche5, line 275: Remove.

n_i2c.c
```diff
@@ -297,3 +207,161 @@ bool i2cNoteReset()
...
+  @param   overflow (out)
+            The contents did not fit inside the provided buffer.
+  @returns  A c-string with an error, or `NULL` if no error ocurred.
+*/
+/**************************************************************************/
+const char *i2cRawReceive(uint8_t *buffer, size_t *size, bool delay, size_t timeoutMs, size_t *overflow)
+{
+    // Check for special case
+    if (!timeoutMs) {
+        timeoutMs = (size_t)-1;
```
haydenroche5, line 233: Why not just check if timeoutMs is zero down in the loop and not do the timeout logic if so?

n_i2c.c
```diff
@@ -297,3 +207,161 @@ bool i2cNoteReset()
...
+        }
+
+        // Exit on timeout
+        if (_GetMs() - startMs >= timeoutMs) {
+            *size = received;
+            buffer[received] = '\0';
+#ifdef ERRDBG
+            _Debug("received only partial reply after timeout:\n");
+            _Debug((char *)buffer);
+            _Debug("^^ partial buffer contents ^^\n");
```
haydenroche5, line 300: Does this need a \n at the start? If I'm reading it correctly, buffer and this partial buffer contents message with the up arrows are going to print on the same line.

n_i2c.c
```diff
@@ -297,3 +207,161 @@ bool i2cNoteReset()
...
+#endif
+            return ERRSTR("transaction incomplete {io}",c_iotimeout);
+        }
+
+        // Delay, simply waiting for the Note to process the request
+        if (delay) {
+            _DelayMs(50);
```
haydenroche5, line 307: I'm trying to connect the 50ms delay here to something in the serial over I2C doc, but I'm not finding it. It would be nice if the two matched up.

n_i2c.c
```diff
@@ -297,3 +207,161 @@ bool i2cNoteReset()
...
+
+/**************************************************************************/
+/*!
+  @brief  Transmit bytes over I2C to the Notecard.
+  @param   buffer
+            A buffer of bytes to transmit.
+  @param   size
+            The count of bytes in the buffer to send
+  @param   delay
+            Respect delay standard transmission delays.
```
haydenroche5, line 323: Reword.

n_serial.c
```diff
@@ -13,66 +13,37 @@
...
-
 /**************************************************************************/
 /*!
-    @brief  Given a JSON string, perform an Serial transaction with the Notecard.
-    @param   json
-               A c-string containing the JSON request object.
-    @param   jsonResponse
-               An out parameter c-string buffer that will contain the JSON
-               response from the Notercard.
+  @brief  Given a JSON string, perform an Serial transaction with the Notecard.
```
haydenroche5, line 18: While we're in here: an Serial -> a serial.

n_serial.c
```diff
@@ -84,60 +55,32 @@ const char *serialNoteTransaction(char *json, char **jsonResponse)
...
     }
 
-    // Allocate a buffer for input, noting that we always put the +1 in the alloc so we can be assured
-    // that it can be null-terminated.  This must be the case because json parsing requires a
-    // null-terminated string.
-    int jsonbufAllocLen = ALLOC_CHUNK;
+    // Allocate a buffer for input, noting that we always put the +1 in the
+    // alloc so we can be assured that it can be null-terminated. This must be
+    // the case because json parsing requires a null-terminated string.
+    size_t jsonbufAllocLen = ALLOC_CHUNK;
```
haydenroche5, line 61: The code from here to line 96 is nearly identical to the corresponding code in i2cNoteTransaction. The only differences are

1. serialRawReceive is used here whereas we use i2cRawReceive in i2cNoteTransaction.
2. i2cNoteTransaction does this:
```
if (overflow > ALLOC_CHUNK) {
    jsonbufAllocLen += overflow;
} else {
    jsonbufAllocLen += ALLOC_CHUNK;
}
```

Is there a reason for the difference in 2? Regarding 1, is this an opportunity to unify the logic into one function used by both i2cNoteTransaction and serialNoteTransaction?

n_serial.c
```diff
@@ -220,3 +162,106 @@ bool serialNoteReset()
     // Done
     return notecardReady;
 }
+
+/**************************************************************************/
+/*!
+  @brief  Receive bytes over Serial to the Notecard.
```
haydenroche5, line 168: to -> from

n_serial.c
```diff
@@ -220,3 +162,106 @@ bool serialNoteReset()
...
+/**************************************************************************/
+/*!
+  @brief  Receive bytes over Serial to the Notecard.
+  @param   buffer
+            A buffer to receive bytes into.
+  @param   size (in/out)
+            - (in) The size of the buffer in bytes.
+            - (out) The length of the received data in bytes.
+  @param   delay
+            Respect delay standard transmission delays.
```
haydenroche5, line 175: Reword.

n_serial.c
```diff
@@ -220,3 +162,106 @@ bool serialNoteReset()
...
+  @param   overflow (out)
+            The overflow amount in bytes.
+  @returns  A c-string with an error, or `NULL` if no error ocurred.
+*/
+/**************************************************************************/
+const char *serialRawReceive(uint8_t *buffer, size_t *size, bool delay, size_t timeoutMs, size_t *overflow)
+{
+    // Check for special case(s)
+    if (!timeoutMs) {
+        timeoutMs = (size_t)-1;
```
haydenroche5, line 188: Same comment I made on the i2c version of this function.

n_serial.c
```diff
@@ -220,3 +162,106 @@ bool serialNoteReset()
...
+    const size_t startMs = _GetMs();
+    for (bool eop = false ; !(*overflow) && !eop ;) {
+        while (!_SerialAvailable()) {
+            if (_GetMs() - startMs >= timeoutMs) {
+                *size = received;
+                buffer[received] = '\0';
+#ifdef ERRDBG
+                _Debug("received only partial reply after timeout:\n");
+                _Debug((char *)buffer);
+                _Debug("^^ partial buffer contents ^^\n");
```
haydenroche5, line 202: Same comment as on i2c. Does this need leading \n?

n_serial.c
```diff
@@ -220,3 +162,106 @@ bool serialNoteReset()
...
+        // Place into the buffer
+        buffer[received++] = ch;
+
+        // Look for end-of-packet marker
+        eop = (ch == '\n');
+
+        // Check overflow condition
+        *overflow = ((received >= *size) && !eop);
+        if (*overflow) {
+            *overflow = ALLOC_CHUNK;
```
haydenroche5, line 221: I think a comment could be helpful here. My interpretation: we don't know how much we have left to read. We just know that the buffer is at least 1 byte too small. So we tell the caller to allocate another chunk of 128, even though that might not be enough either (or it might be more than enough).

n_serial.c
```diff
@@ -220,3 +162,106 @@ bool serialNoteReset()
...
+
+/**************************************************************************/
+/*!
+  @brief  Transmit bytes over Serial to the Notecard.
+  @param   buffer
+            A buffer of bytes to transmit.
+  @param   size
+            The count of bytes in the buffer to send.
+  @param   delay
+            Respect delay standard transmission delays.
```
haydenroche5, line 239: Reword.

n_serial.c
```diff
@@ -220,3 +162,106 @@ bool serialNoteReset()
...
+    }
+
+    // Return it
+    *size = received;
+    return NULL;
+}
+
+/**************************************************************************/
+/*!
+  @brief  Transmit bytes over Serial to the Notecard.
```
haydenroche5, line 233: Serial -> serial

haydenroche5 (changes requested):

n_i2c.c
```diff
@@ -297,3 +207,161 @@ bool i2cNoteReset()
...
+        }
+
+        // If the last byte of the chunk is `\n`, then we have received a
+        // complete message. However, to ensure we pull everything pending from
+        // the Notecard, we will only exit when we've received a newline AND
+        // there's no more bytes available from the Notecard.
+
+        // If there's something available on the Notecard for us to receive, do it
+        if (available > 0) {
+            continue;
```
haydenroche5, line 285: I think we need to check the timeout before we continue here.

haydenroche5 (changes requested):

n_i2c.c
haydenroche5: We also won't have the 50 ms delay if we continue here.

haydenroche5 (changes requested):

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
...
+        return ERRSTR("buffer size exceeds available memory", c_mem);
+    }
+
+    // Calculate MD5
+    char hash_string[NOTE_MD5_HASH_STRING_SIZE] = {0};
+    NoteMD5HashString(data, dataLen, hash_string, NOTE_MD5_HASH_STRING_SIZE);
+
+    // Encode COBS data
+    uint32_t encLen = cobsEncodedLength(data,dataLen);
+    if ((encLen + 1) > bufLen) {
```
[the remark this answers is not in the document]
zfields: kill me now...

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
...
+    // Issue a `"card.binary"` request.
+    rsp = NoteRequestResponse(NoteNewRequest("card.binary"));
+    if (!rsp) {
+        return ERRSTR("unable to validate request", c_err);
+    }
+
+    // Ensure the transaction doesn't return an error
+    // to confirm your binary was received
+    if (NoteResponseError(rsp)) {
+        JDelete(rsp);
```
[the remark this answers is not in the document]
zfields: naw, it's a secret...

n_hooks.c
zfields: This is pattern matching on existing API.

I believe it prevents the problem from it "accidentally working" if you switched away from a given interface.

n_i2c.c
zfields: I didn't want special handling, but I like that better.

n_i2c.c
zfields: This is not documented. This should only happen in the rare occurrence that we have a received a partial message (no newline) and we are awaiting the rest of the message for some reason, but it came back with zero (0) :joy: bytes available.

n_serial.c
```diff
@@ -220,3 +162,106 @@ bool serialNoteReset()
...
+  @param   buffer
+            A buffer of bytes to transmit.
+  @param   size
+            The count of bytes in the buffer to send.
+  @param   delay
+            Respect delay standard transmission delays.
+  @returns  A c-string with an error, or `NULL` if no error ocurred.
+*/
+/**************************************************************************/
+const char *serialRawTransmit(uint8_t *buffer, size_t size, bool delay)
```
m-mcgowan, line 243: It was a bit confusing figuring out the different levels of transmiit. NoteSerialTransmit _SerialTransmit (which are the "raw" transmission hooks.) How about renaming this to `serialChunkedTransmit` or similar?

n_helpers.c
```diff
@@ -74,9 +79,204 @@ static short normalYearDaysByMonth[] = {0, 31, 59, 90, 120, 151, 181, 212, 243,
...
+/*!
+  @brief  Receive a large binary object from the Notecard's binary buffer
+  @param  buffer A buffer to hold the result
+  @param  bufLen The total length of the buffer
+  @returns  NULL on success, else an error string pointer.
+  @note  Buffers are decoded in place, the original contents of the buffer will
+         be modified.
+*/
+/**************************************************************************/
+const char * NoteBinaryReceive(uint8_t * buffer, size_t bufLen)
```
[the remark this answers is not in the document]
haydenroche5: There's a discussion about this in the notecard-fw Slack channel. The Notecard can't do what you're describing today, but I (and Zak) want this feature, too.

n_cobs.c
zfields: This comment isn't even right. This is the point of my whole cry baby fit in Slack this morning.

n_i2c.c
zfields: It's left-over from experimentation. It is a departure from the way we have previously done it. It would allow for allocations larger than `ALLOC_CHUNK`. I'm not sure this is the direction we want to go, because it may cause heap corruption.

I need to have a discussion with Ray, to understand the original intent.

n_i2c.c
zfields: It is compatible with the original implementation. See previous comment.

n_helpers.c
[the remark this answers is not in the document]
zfields: Unfortunately, that is not implemented on the Notecard. Both Hayden and I have independently requested this feature in the #notecard-fw channel.

n_serial.c
zfields: I'm not in love with the name I have chosen.

`serialChunkedTransmit()` sounds reasonable, let me stew on it to ensure it captures the spirit of the function.

n_helpers.c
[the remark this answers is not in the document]
haydenroche5: Is there an echo in here? :laughing:

haydenroche5 (changes requested):

n_helpers.c
zfields: I want to run this by Ray, but I think we need a separate API `NoteBinaryAllocRecieve()`, so it's explicit who's responsible for cleaning up the buffer.

zfields: before I do that, I do want to think through how this may facilitate with Mat's feedback down below. Maybe leaving the door open for a partial reception.

n_serial.c
zfields: I read through the function with fresh eyes, and your name seems to make perfect sense. :+1: 

Thanks for the suggestion!

n_helpers.c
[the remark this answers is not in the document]
m-mcgowan: What can we do to make our case?

m-mcgowan:
I just noticed the style checks failed. We should add a pre-commit hook so that these checks are ran locally before pushing. And while we're at it, might as well run the tests too.

n_helpers.c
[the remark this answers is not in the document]
zfields: Honestly, I don't think this is a priority item when considering all the other goings on. I'm certain it will have it's day, but I imagine it will be customer driven.

That being said, I would make mention of this in #notecard-fw, so there is another vote in the tally.

n_helpers.c
```diff
@@ -116,9 +117,11 @@ const char * NoteBinaryReceive(uint8_t * buffer, size_t bufLen)
     const size_t cobs = JGetInt(rsp,"cobs");
     JDelete(rsp);
     if (!cobs) {
+        _Debug("no data on notecard");
         return ERRSTR("no data on notecard", c_err);
     }
     if (cobs > bufLen) {
+        _Debug("insufficient buffer size");
```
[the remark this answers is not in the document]
zfields: Yes, I thought about that at one point and totally forgot. I will switch them over! :taco:

PR #86, opened 2023-07-23
Optimize I2C and Serial Transmission

zfields:
These changes were discovered when preparing for COBS work.


n_serial.c
```diff
@@ -59,18 +49,18 @@ const char *serialNoteTransaction(char *json, char **jsonResponse)
         }
     }
 
-    // Free the transmit buffer
-    _Free(transmitBuf);
+    // Append newline to the transaction
+    _SerialTransmit((uint8_t *)"\r\n", 2, true);
```
haydenroche5, line 53: Should we use `c_newline` and `c_newline_len` here? Not critical, but the constants are there, so might as well use them.

haydenroche5 (approved):

n_serial.c
zfields: yes we should

zfields:
included in COBS PR.

PR #89, opened 2023-08-17
chore: refactor NoteBinaryTransmit parameter

zfields:
change `append` from a boolean to a size_t to provide a target index into the binary for cross validation


m-mcgowan (approved):
LGTM!

PR #90, opened 2023-08-17
feat: Add NoteBinary* functions for encoding and decoding.

haydenroche5:
And make the existing NoteBinary functions use these instead of the cobs* ones.

### DIFF
--- a/n_helpers.c
+++ b/n_helpers.c
@@ -84,6 +84,99 @@ NOTE_C_STATIC void setTime(JTIME seconds);
 NOTE_C_STATIC bool timerExpiredSecs(uint32_t *timer, uint32_t periodSecs);
 NOTE_C_STATIC int ytodays(int year);
 
+static const char BINARY_EOP = '\n';
+
+//**************************************************************************/
+/*!
+  @brief  Decode a binary payload received from the Notecard.
+  @param  inBuf The binary payload.
+  @param  inLen The length of the binary payload.
+  @param  outBuf The buffer to write the decoded payload to. This can be the
+                 same address as inBuf, allowing for in-place decoding.
+  @param  outLen On input, holds the length of outBuf. On output, holds the
+                 length of the decoded data.
+  @returns  NULL on success, else an error string pointer.
+*/
+/**************************************************************************/
+const char *NoteBinaryDecode(const uint8_t *inBuf, uint32_t inLen,
+                             uint8_t *outBuf, uint32_t *outLen)
+{
+    if (inBuf == NULL || outBuf == NULL || outLen == NULL) {
+        NOTE_C_LOG_ERROR("NULL parameter");
+        return ERRSTR("NULL parameter", c_err);
+    }
+
+    if (*outLen < cobsGuaranteedFit(inLen)) {
+        NOTE_C_LOG_ERROR("output buffer too small");
+        return ERRSTR("output buffer too small", c_err);
+    }
+
+    *outLen = cobsDecode((uint8_t *)inBuf, inLen, BINARY_EOP, outBuf);
+
+    return NULL;
+}
+
+//**************************************************************************/
+/*!
+  @brief  Binary encode a buffer to prepare it for transmission to the Notecard.
+  @param  inBuf The data to encode.
+  @param  inLen The length of the data to encode.
+  @param  outBuf The buffer to write the encoded data to. This can be the
+                 same address as inBuf, allowing for in-place encoding.
+  @param  outLen On input, holds the length of outBuf. On output, holds the
+                 length of the encoded data.
+  @returns  NULL on success, else an error string pointer.
+*/
+/**************************************************************************/
+const char *NoteBinaryEncode(const uint8_t *inBuf, uint32_t inLen,
+                             uint8_t *outBuf, uint32_t *outLen)
+{
+    if (inBuf == NULL || outBuf == NULL || outLen == NULL) {
+        NOTE_C_LOG_ERROR("NULL parameter");
+        return ERRSTR("NULL parameter", c_err);
+    }
+
+    if (*outLen < cobsEncodedMaxLength(inLen)) {
+        if (*outLen < cobsEncodedLength(inBuf, inLen)) {
+            NOTE_C_LOG_ERROR("output buffer too small");
+            return ERRSTR("output buffer too small", c_err);
+        }
+    }
+
+    *outLen = cobsEncode((uint8_t *)inBuf, inLen, BINARY_EOP, outBuf);
+
+    return NULL;
+}
+
+//**************************************************************************/
+/*!
+  @brief  Given an input buffer and its length, compute the exact number of
+          bytes required to binary encode that buffer.
+  @param  buf The buffer to encode.
+  @param  len The length of the buffer.
+  @note   This function iterates over the entire buffer to compute the result.
+          For a constant-time alternative, see NoteBinaryEncodedMaxLength.
+  @returns  The required buffer size to hold the encoded data.
+*/
+/**************************************************************************/
+uint32_t NoteBinaryEncodedLength(const uint8_t *buf, uint32_t len)
+{
+    return cobsEncodedLength(buf, len);
+}
+
+//**************************************************************************/
+/*!
+  @brief  Given an input buffer's length, compute the maximum output buffer size
+          needed to binary encode the input buffer.
+  @param  len The length of the input buffer.
+  @returns  The max required buffer size to hold the encoded data.
+*/
+/**************************************************************************/
+uint32_t NoteBinaryEncodedMaxLength(uint32_t len)
+{
+    return cobsEncodedMaxLength(len);
+}
+
 //**************************************************************************/
 /*!
   @brief  Receive a large binary object from the Notecard's binary buffer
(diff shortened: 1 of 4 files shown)


zfields:
Let's put a pin in the this until we (you, Mat and I) get a chance to talk through the long-term ramifications of the API.

n_helpers.c
```diff
@@ -84,6 +84,45 @@ NOTE_C_STATIC void setTime(JTIME seconds);
...
 NOTE_C_STATIC int ytodays(int year);
 
+static const char BINARY_EOP = '\n';
+
+const char *NoteBinaryDecode(const uint8_t *inBuf, uint32_t inLen,
+                             uint8_t *outBuf, uint32_t *outLen)
+{
+    uint32_t decLen = cobsDecode((uint8_t *)inBuf, inLen, BINARY_EOP, outBuf);
+
+    if (outLen != NULL) {
```
haydenroche5, line 94: In `NoteBinaryTransmit`, we decode the user's input buffer on errors. In those cases, we have no use for the output length, so I deliberately made this function permit `outLen == NULL`.

n_helpers.c
```diff
@@ -348,25 +393,27 @@ const char * NoteBinaryTransmit(uint8_t * data, size_t dataLen, size_t bufLen, s
         return ERRSTR("buffer size exceeds available memory", c_mem);
     }
 
-    // Confirm buffer is large enough to in place encode as COBS
```
haydenroche5, line 351: `NoteBinaryEncode` does this check for us now, so I removed it.

n_helpers.c
```diff
@@ -348,25 +393,27 @@ const char * NoteBinaryTransmit(uint8_t * data, size_t dataLen, size_t bufLen, s
...
     memmove(data + dataShift, data, dataLen);
-    encLen = cobsEncode(data + dataShift, dataLen, '\n', data);
-    data[encLen] = '\n';
+
+    // outLen holds the buffer size available for encoding. The -1 accounts for
+    // one byte of space we need to save for a newline to mark the end of the
+    // packet. When NoteBinaryEncode returns, outLen will hold the encoded
+    // length.
+    uint32_t outLen = bufLen - 1;
+    err = NoteBinaryEncode(data + dataShift, dataLen, data, &outLen);
```
zfields, line 410: This is a reasonable use of intra-interface calls. I see `NoteBinaryTransmit` and `NoteBinaryReceive` as second order functions, and the diff illustrates the value provided.

n_helpers.c
```diff
@@ -389,33 +436,33 @@ const char * NoteBinaryTransmit(uint8_t * data, size_t dataLen, size_t bufLen, s
                 // On errors, we restore the caller's input buffer by COBS
                 // decoding it. The caller is then able to retry transmission
                 // with their original pointer to this buffer.
-                cobsDecode(data, encLen, '\n', data);
+                NoteBinaryDecode(data, outLen, data, NULL);
```
zfields, line 439: Don't love the `NULL` parameter usage here, should be `bufLen`, no?

n_helpers.c
```diff
@@ -376,7 +423,7 @@ const char * NoteBinaryTransmit(uint8_t * data, size_t dataLen, size_t bufLen, s
         // Issue a "card.binary.put"
         J *req = NoteNewRequest("card.binary.put");
         if (req) {
-            JAddIntToObject(req, "cobs", encLen);
+            JAddIntToObject(req, "cobs", outLen);
```
zfields, line 426: I don't understand this rename.

`encLen` (encoded length) seems more expressive in the context of this function. If it was renamed to match the parameters of `NoteBinaryDecode`, then should it not be `inLen`?

n_helpers.c
```diff
@@ -84,6 +84,45 @@ NOTE_C_STATIC void setTime(JTIME seconds);
...
+        *outLen = decLen;
+    }
+
+    return NULL;
+}
+
+const char *NoteBinaryEncode(const uint8_t *inBuf, uint32_t inLen,
+                             uint8_t *outBuf, uint32_t *outLen)
+{
+    if (*outLen < NoteBinaryEncodedMaxLength(inLen)) {
```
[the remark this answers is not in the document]
m-mcgowan: It's just to optimize performance.  The first check is O(1) and is a quick check. The second check is O(N).

n_helpers.c
m-mcgowan: Agreed, andI feel the same applies to all the cases mentioned.

n_helpers.c
m-mcgowan: Yes, if we can ascertain the lower bound on the required length for the decoded data, that's preferable to just trying and hoping it doesn't crash. 

Is there a better name for `cobsGuaranteedFit()`? To me, it's not immediately obvious this is talking about decoding space required, and still wasn't all that clear after I read the doc comments.

n_helpers.c
```diff
@@ -84,6 +84,45 @@ NOTE_C_STATIC void setTime(JTIME seconds);
...
+        if (*outLen < NoteBinaryEncodedLength(inBuf, inLen)) {
```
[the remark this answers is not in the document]
haydenroche5: I'm fine either way, but I'm curious what an example of this problem looks like. Not sure I've seen it myself.

n_helpers.c
```diff
@@ -169,8 +208,12 @@ const char * NoteBinaryReceive(uint8_t * buffer, size_t bufLen,
     // part of the binary payload, so we decrement the length by 1 to remove it.
     --bufLen;
 
+    uint32_t decLen = bufLen;
     // Decode it in-place, which is safe because decoding shrinks
-    const size_t decLen = cobsDecode(buffer, bufLen, '\n', buffer);
+    err = NoteBinaryDecode(buffer, bufLen, buffer, &decLen);
```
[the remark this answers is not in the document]
haydenroche5: - If we change the codec to something other than COBS, then we don't have to change this code again. It just works.
- We get test coverage of the new functions "for free."
- We are using the same functions internally that our users will be using externally, which makes us more confident that they actually work/are bug-free.

I'm not opposed to switching over to the cobs* functions, but I'm not seeing the benefit to doing that, and like I commented above, I'm not sure what an example of the problem you're talking about looks like in practice.

n_helpers.c
haydenroche5: https://github.com/blues/note-c/pull/90#discussion_r1304978894

n_helpers.c
haydenroche5: It's an in/out parameter now. On input, it's the length of the output buffer, so encLen didn't feel like a great name anymore. I felt outLen kind of captured the semantics of both input and output: on input, it's the length of the output buffer, and on output, it's the length of the actual data in the output buffer.

n_helpers.c
zfields: I agree, but that is Ray's naming convention and it is taken directly from the Notecard firmware. My goal is to keep these files identical.

Changing the name is a reasonable request, but it should be done outside of this PR.

n_helpers.c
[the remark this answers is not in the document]
zfields: Cool, that sounds great. I completely overlooked that.

n_helpers.c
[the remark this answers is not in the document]
zfields: Personally, I see as two levels of API. The level of API that facilitate the direct usage of the Notecard API (levels 2 and 3), and the level  that completely abstracts it (level 1).

I really have to squint to see all three levels and not just two, but I do understand how you're making the distinction.

I think you are being optimistic that we wouldn't need to change all the implementations, but I don't have an overly strong opinion here.

n_helpers.c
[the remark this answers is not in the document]
zfields: Mat did a really good job of breaking down the "levels" of the API.

> 1.  NoteBinaryTransmit/Receive and the helpers for required Transmit/Receive buffer sizes.
> 1.  NoteBinaryEncode/Decode/[Length]
> 1.  NoteBinaryMaxEncodedLength

When you co-mingle in this fashion, you are assuming that all engineers that touch this code will be able to make the same break down in their head.

haydenroche5:
I think I addressed all the review comments. @zfields and I discussed in a Slack huddle and settled on using the cobs* functions internally for the new functions. For NoteBinaryReceive/Transmit, we use the new functions. Latest push also adds Doxygen comments.

n_helpers.c
```diff
@@ -84,6 +84,89 @@ NOTE_C_STATIC void setTime(JTIME seconds);
...
+                 same address as inBuf, allowing for in-place encoding.
+  @param  outLen On input, holds the length of outBuf. On output, holds the
+                 length of the encoded data.
+  @returns  NULL on success, else an error string pointer.
+*/
+/**************************************************************************/
+const char *NoteBinaryEncode(const uint8_t *inBuf, uint32_t inLen,
+                             uint8_t *outBuf, uint32_t *outLen)
+{
+    if (*outLen < cobsEncodedMaxLength(inLen)) {
```
zfields, line 129: It's dangerous to dereference a pointer (especially on a public API) without first checking it for `NULL`.

n_helpers.c
```diff
@@ -348,25 +437,27 @@ const char * NoteBinaryTransmit(uint8_t * data, size_t dataLen, size_t bufLen, s
...
     const size_t dataShift = (bufLen - dataLen);
     memmove(data + dataShift, data, dataLen);
-    encLen = cobsEncode(data + dataShift, dataLen, '\n', data);
-    data[encLen] = '\n';
+
+    // outLen holds the buffer size available for encoding. The -1 accounts for
+    // one byte of space we need to save for a newline to mark the end of the
+    // packet. When NoteBinaryEncode returns, outLen will hold the encoded
+    // length.
+    uint32_t outLen = bufLen - 1;
```
zfields, line 453: > // outLen holds the buffer size available for encoding.

This comment is only necessary, because you have named the variable to match the parameter (context) of the function you are calling into, as opposed to the context of the current function in which it is being used.

zfields (changes requested):

n_helpers.c
haydenroche5: I agree. I just don't see us coding defensively like that in note-c, historically-speaking, but I think it would be good to start. I'll add checks.

n_helpers.c
haydenroche5: Oh yeah I meant to change this back to `encLen` per our conversation.

zfields (approved):

PR #93, opened 2023-08-31
chore: support binary chunked receive

zfields:


zfields:
[Reimplement `NoteBinaryReceive` with chunked receive in `note-c`](https://trello.com/c/oJWyNL24/186-reimplement-notebinaryreceive-with-chunked-receive-in-note-c)

n_helpers.c
```diff
@@ -182,28 +182,33 @@ uint32_t NoteBinaryEncodedMaxLength(uint32_t len)
...
     }
-    if (!requiredRxBufLen) {
-        NOTE_C_LOG_ERROR("no data on notecard");
-        return ERRSTR("no data on notecard", c_err);
+    if (!dataLen) {
+        NOTE_C_LOG_ERROR("dataLen cannot be NULL");
+        return ERRSTR("dataLen cannot be NULL", c_err);
     }
-    if (requiredRxBufLen > bufLen) {
+    if (bufLen < (NoteBinaryEncodedMaxLength(*dataLen) + sizeof('\n'))) {
```
m-mcgowan, line 211: Why not use `NoteBinaryRequiredBuffer()` as described in the docs?

I'm using NoteBinaryRequiredBuffer from the `cobs` response but it's still telling me the buffer is too small. I changed `(NoteBinaryEncodedMaxLength(*dataLen) + sizeof('\n'))` to `NoteBinaryRequiredBuffer()` and it gets past that check, but later fails with

```
{"req":"card.binary.get","offset":0,"length":73,"crc":"0053:78FA2A08"}
{"err":"offset+length extends beyond what's in the binary I/O buffer"}
.pio/libdeps/debug/Blues Wireless Notecard/src/note-c/n_helpers.c:229 (ERROR) offset+length extends beyond what's in the binary I/O buffer
```
I assume then that the `cobs` response on `web.put` requests is the COBS-encoded size of the buffer? But then how do I determine the size of the actual data?

zfields: It was out of my usual hesitation to call into the same API.

However, doing so helps to eliminate the bug (i.e. `\n` is interpreted as an `int` and `sizeof()` returns 4 instead of 1), so I changed my opinion.

n_helpers.c
```diff
@@ -98,8 +140,8 @@ static const char BINARY_EOP = '\n';
   @returns  NULL on success, else an error string pointer.
 */
 /**************************************************************************/
-const char *NoteBinaryDecode(const uint8_t *inBuf, uint32_t inLen,
-                             uint8_t *outBuf, uint32_t *outLen)
+const char * NoteBinaryDecode(const uint8_t *inBuf, uint32_t inLen,
```
haydenroche5, line 143: Preposterous!

n_helpers.c
```diff
@@ -390,17 +442,23 @@ const char * NoteBinaryReset(void)
...
+  @note  You may use `NoteBinaryRequiredBuffer()` to calculate the required size
          for the buffer pointed to by the `data` parameter.
 */
 /**************************************************************************/
 const char * NoteBinaryTransmit(uint8_t * data, size_t dataLen, size_t bufLen,
                                 size_t offset)
 {
+    // Validate parameter(s)
+    if (!dataLen) {
+        NOTE_C_LOG_ERROR("dataLen cannot be NULL");
```
haydenroche5, line 458: dataLen isn't a pointer.

Also, should we validate the other parameters, to be consistent?

n_helpers.c
zfields: I did make this change, then in a subsequent commit I deleted the API it was relying upon, so I ended up fixing the bug and changing it back.

n_helpers.c
```diff
@@ -86,20 +86,154 @@ NOTE_C_STATIC int ytodays(int year);
...
+        // 1 byte for the terminating newline.
+        *len = cobs + 1;
+    }
+
+    return NULL;
+}
+
+//**************************************************************************/
+/*!
+  @brief  Reset the Notecard's binary buffer
```
haydenroche5, line 189: Missing period.

n_helpers.c
```diff
@@ -86,20 +86,154 @@ NOTE_C_STATIC int ytodays(int year);
...
+
 //**************************************************************************/
 /*!
   @brief  Decode a binary payload received from the Notecard.
+
   @param  inBuf The binary payload.
   @param  inLen The length of the binary payload.
   @param  outBuf The buffer to write the decoded payload to. This can be the
-                 same address as inBuf, allowing for in-place decoding.
+                 same address as inBuf, allowing for in place decoding.
```
haydenroche5, line 228: Shouldn't this be in-place, as it was before?

n_helpers.c
```diff
@@ -118,18 +252,21 @@ const char *NoteBinaryDecode(const uint8_t *inBuf, uint32_t inLen,
...
 //**************************************************************************/
 /*!
+
   @brief  Binary encode a buffer to prepare it for transmission to the Notecard.
+
   @param  inBuf The data to encode.
   @param  inLen The length of the data to encode.
   @param  outBuf The buffer to write the encoded data to. This can be the
-                 same address as inBuf, allowing for in-place encoding.
+                 same address as inBuf, allowing for in place encoding.
```
haydenroche5, line 261: Same comment as above.

n_helpers.c
```diff
@@ -150,60 +287,87 @@ const char *NoteBinaryEncode(const uint8_t *inBuf, uint32_t inLen,
...
-  @param  len The length of the buffer.
-  @note   This function iterates over the entire buffer to compute the result.
-          For a constant-time alternative, see NoteBinaryEncodedMaxLength.
-  @returns  The required buffer size to hold the encoded data.
-*/
+  @brief  Compute the maximum decoded data length guaranteed
+          to fit into a fixed-size buffer, after being encoded.
+
+          This API is designed for a space constrained environment, where a
+          working buffer has been allocated to facilitate with binary
```
haydenroche5, line 294: facilitate with -> facilitate

n_helpers.c
```diff
@@ -150,60 +287,87 @@ const char *NoteBinaryEncode(const uint8_t *inBuf, uint32_t inLen,
...
 
 //**************************************************************************/
 /*!
   @brief  Receive a large binary object from the Notecard's binary buffer
+
   @param  buffer A buffer to hold the binary object
   @param  bufLen The total length of the provided buffer
-  @param  dataLen An out parameter to hold the length of the decoded data in the
-                  output buffer.
+  @param  decodedOffset The offset to the decoded binary data already residing
```
haydenroche5, line 341: I thought you were gonna go with `notecardOffset`.

n_helpers.c
```diff
@@ -150,60 +287,87 @@ const char *NoteBinaryEncode(const uint8_t *inBuf, uint32_t inLen,
...
   @returns  NULL on success, else an error string pointer.
-  @note  Buffers are decoded in place. The original contents of the buffer
-         will be modified.
-  @note  To determine the necessary size for the buffer, use
-         NoteBinaryRequiredRxBuffer.
-*/
+
+  @note  The buffer must be large enough to hold the encoded value of the
+         data store contents from the requested offset for the specified length.
+  @note  To determine the necessary buffer size for a given data length, use
```
haydenroche5, line 353: This note is sort of just a continuation of the previous one. I'd just consolidate it into one note. I'd also explain that the +1 below if or the newline. Also, `NoteBinaryMaxEncodedLength` takes a parameter, so I'd change `NoteBinaryMaxEncodedLength() + 1` to `NoteBinaryMaxEncodedLength(decodedLen) + 1`.

n_helpers.c
```diff
@@ -150,60 +287,87 @@ const char *NoteBinaryEncode(const uint8_t *inBuf, uint32_t inLen,
...
     }
-    if (!requiredRxBufLen) {
-        NOTE_C_LOG_ERROR("no data on notecard");
-        return ERRSTR("no data on notecard", c_err);
+    if (!decodedLen) {
+        NOTE_C_LOG_ERROR("decodedLen cannot be NULL");
+        return ERRSTR("decodedLen cannot be NULL", c_err);
     }
-    if (requiredRxBufLen > bufLen) {
+    if (bufLen < (cobsEncodedMaxLength(*decodedLen) + sizeof(char))) {
```
haydenroche5, line 370: In the doc above we say +1 but here we do + `sizeof(char)`. I think either is fine (slight preference for +1 -- it's easier to read and is there really a platform where `sizeof(char)` is not 1?), but I would be consistent.

n_helpers.c
```diff
@@ -263,13 +425,13 @@ const char * NoteBinaryReceive(uint8_t * buffer, size_t bufLen,
     --bufLen;
 
     uint32_t decLen = bufLen;
-    // Decode it in-place, which is safe because decoding shrinks
+    // Decode it in place, which is safe because decoding shrinks
```
haydenroche5, line 428: in-place is right: https://en.wikipedia.org/wiki/In-place_algorithm

n_helpers.c
```diff
@@ -290,117 +452,34 @@ const char * NoteBinaryReceive(uint8_t * buffer, size_t bufLen,
...
-    } else {
-        // Otherwise, the required length is cobs + 1: the binary data plus
-        // 1 byte for the terminating newline.
-        *size = cobs + 1;
-    }
+  @brief  Transmit a large binary object to the Notecard's binary buffer
 
-    return NULL;
-}
+  @param  unencodedData  A buffer with data to encode in place
```
haydenroche5, line 457: in-place

n_helpers.c
```diff
@@ -290,117 +452,34 @@ const char * NoteBinaryReceive(uint8_t * buffer, size_t bufLen,
...
-  @brief  Transmit a large binary object to the Notecard's binary buffer
-  @param  data  A buffer with data to encode in place
-  @param  dataLen  The length of the data in the buffer
-  @param  bufLen  The total length of the buffer
-  @param  offset  The offset where the `data` buffer should be appended to the
-                  unencoded binary data already residing on the Notecard. This
-                  does not provide random access, but rather ensures alignment
-                  between the callers expectation and Notecard.
-  @returns  NULL on success, else an error string pointer.
   @note  Buffers are encoded in place, the buffer _MUST_ be larger than the data
```
haydenroche5, line 467: in-place -- I'll stop saying it :)

n_helpers.c
```diff
@@ -435,39 +514,40 @@ const char * NoteBinaryTransmit(uint8_t * data, size_t dataLen, size_t bufLen,
     // Validate the index provided by the caller, against the `length` value
     // returned from the Notecard to ensure the caller and Notecard agree on
     // how much data is residing on the Notecard.
-    if (offset != len) {
+    if ((long)notecardOffset != len) {
```
haydenroche5, line 517: Is is theoretically possible for this to get us into trouble? Thinking about if notecardOffset has the MSB set and long is 32-bits.

n_helpers.c
```diff
@@ -435,39 +514,40 @@ const char * NoteBinaryTransmit(uint8_t * data, size_t dataLen, size_t bufLen,
...
         NOTE_C_LOG_ERROR("notecard data length is misaligned with offset");
         return ERRSTR("notecard data length is misaligned with offset", c_mem);
     }
 
-    // When `offset` is zero, the entire buffer is available
-    const size_t remaining = (offset ? (max - len) : max);
-    if (dataLen > remaining) {
+    // When offset is zero, the entire buffer is available
```
haydenroche5, line 522: There are a lot of buffers being thrown around. I'd specify that by "the entire buffer", we mean "the entire binary buffer on the Notecard."

n_helpers.c
```diff
@@ -435,39 +514,40 @@ const char * NoteBinaryTransmit(uint8_t * data, size_t dataLen, size_t bufLen,
...
+    NoteMD5HashString(unencodedData, unencodedLen, hashString, NOTE_MD5_HASH_STRING_SIZE);
 
     // Shift the data to the end of the buffer. Next, we'll encode the data,
     // outputting the encoded data to the front of the buffer.
-    const size_t dataShift = (bufLen - dataLen);
-    memmove(data + dataShift, data, dataLen);
+    const uint32_t dataShift = (bufLen - unencodedLen);
+    memmove(unencodedData + dataShift, unencodedData, unencodedLen);
 
     // outLen holds the buffer size available for encoding. The -1 accounts for
```
haydenroche5, line 538: It's - sizeof(char) now, rather than -1.

n_helpers.c
```diff
@@ -435,39 +514,40 @@ const char * NoteBinaryTransmit(uint8_t * data, size_t dataLen, size_t bufLen,
...
     // one byte of space we need to save for a newline to mark the end of the
     // packet. When NoteBinaryEncode returns, outLen will hold the encoded
     // length.
-    uint32_t encLen = bufLen - 1;
-    err = NoteBinaryEncode(data + dataShift, dataLen, data, &encLen);
+    uint32_t encLen = (bufLen - sizeof(char));
```
haydenroche5, line 542: Truly doesn't matter, but why use parentheses here?

haydenroche5 (changes requested):
Reviewing tests next, but wanted to get my source code comments out first.

haydenroche5 (commented):
Tests look good.

n_helpers.c
zfields: `len` comes from `JGetInt` which returns a long. So we would be in trouble either way.

If the underlying hardware were able to support > 2MB of flash, then we would immediately have a problem with CJSON, so the first signs of fire would not be at the level of this API.

n_helpers.c
zfields: https://ell.stackexchange.com/a/236898

I went looking before, but now I found the definitive answer, and yes it should be hyphenated, because it is an adjective.

The following demonstrate the correct usage:

- "in-place decoding"
- "decoding in place"

This is crazy to me :exploding_head:!

n_helpers.c
zfields: not this one.

n_helpers.c
zfields: it is a habit I have for anytime there is a compound expression.

If the expression were to grow, then you must consider the parenthesis instead of just carelessly appending.

For example, imagine you wanted to double the size.

:point_down: oops!
```c
uint32_t encLen = bufLen - sizeof(char) * 2;
```
```c
uint32_t encLen = (bufLen - sizeof(char)) * 2;
```
:point_up_2: :+1:

n_helpers.c
zfields: That was for `NoteBinaryTransmit()`. It is redundant on a receive, and I believe it is more important to understand that data on the Notecard is in a decoded state when trying to calculate offsets and lengths.

n_helpers.c
zfields: not here.

n_helpers.c
zfields: only adjectives

n_helpers.c
zfields: Just went with +1

haydenroche5 (approved):

PR #94, opened 2023-09-08
chore: eliminate [in/out] parameters

zfields:


n_helpers.c
```diff
@@ -226,26 +226,30 @@ const char * NoteBinaryDataReset(void)
...
+                              uint32_t *decLen)
 {
-    if (inBuf == NULL || outBuf == NULL || outLen == NULL) {
+    if (inBuf == NULL || outBuf == NULL || decLen == NULL) {
         NOTE_C_LOG_ERROR("NULL parameter");
         return ERRSTR("NULL parameter", c_err);
     }
 
-    if (*outLen < cobsGuaranteedFit(inLen)) {
+    if (outLen < cobsGuaranteedFit(inLen)) {
```
m-mcgowan, line 247: `cobsGuaranteedFit` adds the +1 for the transport framing. I think we can make things more consistent, by making a clear separation between the pure COBS encoding requirements and the transport framing requirements (the +1.). Adding `#define COBS_TRANSPORT_OVERHEAD=1` will make this easier to spot in the code.

n_helpers.c
```diff
@@ -331,19 +340,63 @@ uint32_t NoteBinaryMaxEncodedLength(uint32_t unencodedLength)
     return cobsEncodedMaxLength(unencodedLength);
```
m-mcgowan, line 340: Needs a +1 for the framing overhead. 

We can avoid littering the code with +1 everywhere, with 

```c++
cobsEncodedToBufferLength(uint32_t encodedLength) {
   return encodedLength+COBS_TRANSPORT_OVERHEAD;
}
```

That DRYs it up and means just a single change should the transport framing change, although here the DRY and clear semantics is more pertinent atm.

The code would use this whenever it needs to convert from the cobs encoding data length (such as the `cobs` value from the Notecard, or the max encoded length) to the actual buffer size needed to receive it.

n_helpers.c
```diff
@@ -354,21 +407,24 @@ uint32_t NoteBinaryMaxEncodedLength(uint32_t unencodedLength)
...
-        NOTE_C_LOG_ERROR("NULL buffer");
-        return ERRSTR("NULL buffer", c_err);
+        const char *err = ERRSTR("NULL buffer", c_bad);
+        NOTE_C_LOG_ERROR(err);
+        return err;
     }
-    if (!decodedLen) {
-        NOTE_C_LOG_ERROR("decodedLen cannot be NULL");
-        return ERRSTR("decodedLen cannot be NULL", c_err);
+    if (bufLen < (cobsEncodedMaxLength(decodedLen) + 1)) {
```
m-mcgowan, line 419: I think that when we want to know about the required size of the buffer we should use `NoteBinaryMaxEncodedLength()` since that's what we expect the callers to use. 

Currently `NoteBinaryMaxEncodedLength` doesn't do the +1 so this call fails at present.

I thought we'd agreed on this last week that it's ok to call NoteBinaryMaxEncodedLength rather than replicating the code, especially when it's about validating buffer sizes.

(This is the 3rd time this has bit us btw.)

The cobsEncodedMaxLength() etc functions are suitable when talking about purely cobs conversion, and shouldn't make adjustments for the framing overhead.

m-mcgowan (changes requested):

n_helpers.c
zfields: I like this idea. To follow the common naming convention, I will use the symbol `COBS_EOP_OVERHEAD`.

From Wikipedia:
```
   [OHB]                                : Overhead byte (Start of frame)
     3+ -------------->|                : Points to relative location of first zero symbol
                       2+-------->|     : Is a zero data byte, pointing to next zero symbol
                                  [EOP] : Location of end-of-packet zero symbol.
     0     1     2     3     4    5     : Byte Position
     03    11    22    02    33   00    : COBS Data Frame
           11    22    00    33         : Extracted Data
     
OHB = Overhead Byte (Points to next zero symbol)
EOP = End Of Packet
```

n_helpers.c
zfields: now promoted up through `cobsEncodedMaxLength()`

n_helpers.c
zfields: I don't understand how the code you're referencing (line 419) failed; it incorporates the `+1`.

Assuming what is currently sitting on line 419 works, it will be resolved by the underlying change to `cobsEncodedMaxLength()`.

m-mcgowan: NoteBinaryMaxEncodedLength didn't add the +1, so it was returning 1 byte less. I feel adding +1 to cobsXXX functions is mixing concerns.

n_helpers.c
```diff
@@ -84,205 +84,89 @@ NOTE_C_STATIC void setTime(JTIME seconds);
...
 //**************************************************************************/
 /*!
-  @brief  Get the length of the data stored on the Notecard. If there's no data
-          stored on the Notecard, then `*len` will return 0.
+  @brief  Decode binary data received from the Notecard.
 
-  @param  len [out] the length of the decoded contents of the Notecard's binary
-          data store.
+  @param  encData The encoded binary data to decode.
+  @param  encLen The length of the encoded binary data.
```
haydenroche5, line 94: Should we change `encLen` to `encDataLen`, to match how we use `decData` and `decDataLen` for `NoteBinaryCodecEncode`?

n_helpers.c
```diff
@@ -84,205 +84,89 @@ NOTE_C_STATIC void setTime(JTIME seconds);
...
-
-  @param  inBuf The binary payload.
-  @param  inLen The length of the binary payload.
-  @param  outBuf The buffer to write the decoded payload to. This can be the
-                 same address as inBuf, allowing for in-place decoding.
-  @param  outLen On input, holds the length of outBuf. On output, holds the
-                 length of the decoded data.
+  @param  decData The decoded binary data to encode.
+  @param  decDataLen The length of the decoded binary data.
+  @param  encBuf The target buffer for the encoded data. This can be in the same
```
haydenroche5, line 134: This can be in the same - > This can be the same

n_helpers.c
```diff
@@ -84,205 +84,89 @@ NOTE_C_STATIC void setTime(JTIME seconds);
...
+                 buffer as `decData`, allowing for in-place encoding (see note).
+  @param  encBufSize The size of `encBuf`.
 
   @returns  NULL on success, else an error string pointer.
```
haydenroche5, line 138: This is outdated.

n_helpers.c
```diff
@@ -326,49 +210,143 @@ uint32_t NoteBinaryMaxDecodedLength(uint32_t bufferSize)
...
+    return NULL;
+}
+
+//**************************************************************************/
+/*!
+  @brief  Get the required buffer length to receive the entire binary object
+          stored on the Notecard.
+
+  @param  len [out] The length required to hold the entire contents of the
+           Notecard's binary data store. If there's no data stored on the
```
haydenroche5, line 270: 270-271 are indented by 1 additional space. This is really unacceptable, @zfields.

n_helpers.c
```diff
@@ -326,49 +210,143 @@ uint32_t NoteBinaryMaxDecodedLength(uint32_t bufferSize)
...
+    return NULL;
+}
+
 //**************************************************************************/
 /*!
   @brief  Receive a large binary object from the Notecard's binary buffer
 
-  @param  buffer A buffer to hold the binary object
-  @param  bufLen The total length of the provided buffer
+  @param  buffer        A buffer to hold the binary range
```
haydenroche5, line 316: I'm good with either convention, but just noting that we aren't aligning the param descriptions for some of the other APIs in this PR. Also, prefer consistency with ending descriptions with a period.

n_helpers.c
```diff
@@ -326,49 +210,143 @@ uint32_t NoteBinaryMaxDecodedLength(uint32_t bufferSize)
...
+    if (bufLen < cobsEncodedMaxLength(decodedLen)) {
+        const char *err = ERRSTR("insufficient buffer size", c_bad);
+        NOTE_C_LOG_ERROR(err);
+        return err;
     }
-    if (bufLen < (cobsEncodedMaxLength(*decodedLen) + 1)) {
-        NOTE_C_LOG_ERROR("insufficient buffer size");
-        return ERRSTR("insufficient buffer size", c_err);
+    if (decodedLen == 0) {
+        const char *err = ERRSTR("decodedLen cannot be zero (0)", c_bad);
```
haydenroche5, line 347: You know I can't resist pointing out "zero (0)". :)

n_helpers.c
```diff
@@ -441,14 +423,51 @@ const char * NoteBinaryReceive(uint8_t * buffer, uint32_t bufLen,
...
+        return err;
     }
 
     // Return `NULL` if success, else error string pointer
     return NULL;
 }
 
+//**************************************************************************/
+/*!
+  @brief  Reset the Notecard's binary buffer.
```
haydenroche5, line 437: Should we call it the binary data store / binary store instead of binary buffer? Same question applies elsewhere where we use the term binary buffer.

n_helpers.c
```diff
@@ -441,14 +423,51 @@ const char * NoteBinaryReceive(uint8_t * buffer, uint32_t bufLen,
...
+            NOTE_C_LOG_ERROR(err);
+            return err;
+        }
+    } else {
+        const char *err = ERRSTR("unable to allocate request", c_mem);
+        NOTE_C_LOG_ERROR(err);
+        return err;
+    }
+
+    return NULL;
```
haydenroche5, line 468: Need to free rsp. Should be caught by the unit test with valgrind.

n_helpers.c
```diff
@@ -564,19 +596,21 @@ const char * NoteBinaryTransmit(uint8_t *unencodedData, uint32_t unencodedLen,
 
             // Ensure the transaction doesn't return an error.
             if (!NoteRequest(req)) {
-                NOTE_C_LOG_ERROR("failed to initialize binary transaction");
+                const char *err = ERRSTR("failed to initialize binary transaction", c_err);
+                NOTE_C_LOG_ERROR(err);
                 _UnlockNote();
                 // On errors, we restore the caller's input buffer by COBS
```
haydenroche5, line 602: Might make sense to remove "COBS" here, while you're making changes.

haydenroche5 (changes requested):

n_helpers.c
zfields: :grimacing:

n_helpers.c
zfields: good catch!

n_helpers.c
zfields: I don't understand. What is outdated?

n_helpers.c
zfields: I deliberately chose to add "in". The address MUST be different, but the addresses can be part of (or "in") the same buffer.

n_helpers.c
zfields: I understand the concern, but I don't have the freedom to modify `cobsGuaranteedFit()`.

`cobsEncodedMaxLength()` was designed to be an analog to `cobsGuaranteedFit()`, and I believe it needs to follow the existing precedent.

n_helpers.c
zfields: I reworded it to be more clear and explicit regarding the concern expressed above.

haydenroche5 (approved):

PR #97, opened 2023-09-13
chore: Improve error logging

zfields:


n_request.c
```diff
@@ -426,7 +426,7 @@ J *NoteTransaction(J *req)
             _Free(responseJSON);
             errStr = "crc error {io}";
             lastRequestRetries++;
-            _Debugln("CRC error on response");
+            NOTE_C_LOG_ERROR(ERRSTR("CRC error on response", iobad));
```
haydenroche5, line 429: Should be c_iobad.

haydenroche5 (changes requested):

n_request.c
```diff
@@ -522,9 +523,9 @@ J *noteTransactionShouldLock(J *req, bool lockNotecard)
     // Debug
     if (suppressShowTransactions == 0) {
         if (responseJSON[strlen(responseJSON)-1] == '\n') {
-            _Debug(responseJSON);
+            NOTE_C_LOG_INFO(responseJSON);
         } else {
-            _Debugln(responseJSON);
+            NOTE_C_LOG_INFO(responseJSON);
```
haydenroche5, line 528: This else can be removed.

haydenroche5 (changes requested):

n_request.c
zfields: looks more like the `if` should be removed.

haydenroche5 (approved):

PR #99, opened 2023-09-13
fix: Fix deadlock issue with NoteBinaryReceive/Transmit.

haydenroche5:


n_request.c
```diff
@@ -313,7 +313,23 @@ char *NoteRequestResponseJSON(char *reqJSON)
 /**************************************************************************/
 J *NoteTransaction(J *req)
 {
+    return NoteTransactionShouldLock(req, true);
```
haydenroche5, line 316: @rayozzie I went with `NoteTransactionShouldLock` instead of `uNoteTransaction`. That allows `NoteTransaction` to wrap the new function like so.

rayozzie: Hayden, at a minimum please lower-case the first letter and make sure it isn't in note.h.

I know I was very inconsistent about this early-on, but in all C code I've now completely embraced the Golang notion that private functions and variables always begin lower-cased, and all public functions and variables begin with uppercase.

(If there are other note-c cleanups you want to do to make sure this convention is followed, I'd encourage it.)

The "u" is a separate convention that we are rigorous about in both our hub and notecard repo, but it isn't critical to follow this in note-c.

zfields (approved):

n_request.c
zfields: We should do this we EVERYTHING in `note-c`, so it means something throughout the library.

PR #100, opened 2023-09-13
chore: Refactor timeout logic

zfields:


n_serial.c
```diff
@@ -216,10 +216,17 @@ const char *serialChunkedReceive(uint8_t *buffer, uint32_t *size, bool delay, si
 #endif
                 return ERRSTR("timeout: transaction incomplete {io}",c_iotimeout);
             }
-            if (delay) {
+            // Be socialable while awaiting first byte
```
haydenroche5, line 219: I don't know what this means.

haydenroche5 (changes requested):

haydenroche5 (approved):

n_serial.c
```diff
@@ -76,7 +76,7 @@ const char *serialNoteTransaction(char *request, char **response)
         uint32_t jsonbufAvailLen = (jsonbufAllocLen - jsonbufLen);
 
         // Append into the json buffer
-        const char *err = serialChunkedReceive((uint8_t *)(jsonbuf + jsonbufLen), &jsonbufAvailLen, true, (NOTECARD_TRANSACTION_TIMEOUT_SEC * 1000), &available);
+        const char *err = serialChunkedReceive((uint8_t *)(jsonbuf + jsonbufLen), &jsonbufAvailLen, true, (NOTECARD_INTRA_TRANSACTION_TIMEOUT_SEC * 1000), &available);
```
haydenroche5, line 79: Should this be NOTECARD_INTER_TRANSACTION_TIMEOUT_SEC? It was 30 before. Now it's 1.

n_i2c.c
```diff
@@ -144,7 +144,7 @@ const char *i2cNoteTransaction(char *request, char **response)
         uint32_t jsonbufAvailLen = (jsonbufAllocLen - jsonbufLen);
 
         // Append into the json buffer
-        const char *err = i2cChunkedReceive((uint8_t *)(jsonbuf + jsonbufLen), &jsonbufAvailLen, true, (NOTECARD_TRANSACTION_TIMEOUT_SEC * 1000), &available);
+        const char *err = i2cChunkedReceive((uint8_t *)(jsonbuf + jsonbufLen), &jsonbufAvailLen, true, (NOTECARD_INTRA_TRANSACTION_TIMEOUT_SEC * 1000), &available);
```
haydenroche5, line 147: Same question I asked on the serial version of this.

haydenroche5 (changes requested):

n_serial.c
zfields: The `NOTECARD_INTER_TRANSACTION_TIMEOUT` is satisfied above, then it comes down here and this should actually be the `NOTECARD_INTRA_TRANSACTION_TIMEOUT`.

n_i2c.c
```diff
@@ -112,11 +114,36 @@ const char *i2cNoteTransaction(char *request, char **response)
...
+    // `web.*` transactions are unpredictable because we do not know the
+    // associated payload size or the network conditions. The default timeout
+    // for a `web.*` transaction is 90-seconds. Therefore a perfect functioning
+    // transaction could exceed `NOTECARD_INTER_TRANSACTION_TIMEOUT_SEC`, and
+    // an error would be returned to the caller when no error condition exists.
+    // By interrogating the request, we can discover if the request is a
+    //`web.*` transaction. If so, we will update the timeout to the default
+    // 90-seconds, or the value of the `seconds` parameter specified by the
+    // caller (who better understands context of the `web.*` transaction).
+    size_t transaction_timeout_seconds = NOTECARD_INTER_TRANSACTION_TIMEOUT_SEC;
```
haydenroche5, line 126: Just an idea, but could we move this logic up to the level of `_Transaction` so that we don't have to repeat it for both serial and I2C? You would also have access to the `J *` request rather than having to deal with raw strings, which would make it easier to pull out the `seconds` parameter, if it exists. You'd then pass down the transaction timeout seconds as a new parameter to `i2cNoteTransaction`/`serialNoteTransaction`.

haydenroche5: Also, mixing snake_case and camelCase.

n_i2c.c
```diff
@@ -328,6 +355,12 @@ const char *i2cChunkedReceive(uint8_t *buffer, uint32_t *size, bool delay, size_
         // Add requested bytes to received total
         received += requested;
 
+        // Once we've received any character, we will no longer wait patiently
+        if (requested != 0) {
+            timeoutMs = (NOTECARD_INTRA_TRANSACTION_TIMEOUT_SEC * 1000);
```
haydenroche5, line 360: When `i2cChunkedReceive` is called in `i2cNoteTransaction`, it is called with `timeoutMs` already set to 1 second. The other place it gets called is `NoteBinaryStoreReceive`, from what I can tell, and it's also called with 1 second there. So what's the point of setting it to 1 second here?

haydenroche5 (changes requested):

n_lib.h
```diff
@@ -28,7 +28,8 @@ extern "C" {
     @brief  How long to wait for the card for any given transaction.
 */
 /**************************************************************************/
-#define NOTECARD_TRANSACTION_TIMEOUT_SEC     30
+#define NOTECARD_INTER_TRANSACTION_TIMEOUT_SEC 30
```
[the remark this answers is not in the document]
zfields: Not a bad idea. I was just following precedent. It is intended to be internal, so I don't think it needs a prefix.

n_i2c.c
zfields: death to camels!

haydenroche5 (approved):

n_i2c.c
zfields: After talking it through, we decided the timeout parameter for `_ChunkedTransaction()` is redundant. However, at this time, we did not find value in removing it. As it is not the subject of this PR.

n_i2c.c
zfields: Camels stomp snakes. :frowning_face:

n_lib.h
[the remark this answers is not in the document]
zfields: These defines are not external, so we determined that the `NOTE_C_` prefix was unnecessary. However, we noticed that all the other defines in this file begin with `CARD_` and not `NOTECARD_` so we followed that pattern.

haydenroche5 (approved):

haydenroche5 (approved):

PR #101, opened 2023-09-18
fix: `i2cNoteReset()` logic

zfields:


n_i2c.c
```diff
@@ -212,69 +212,105 @@ const char *i2cNoteTransaction(char *request, char **response, size_t timeoutMs)
...
+        //       versions of the Notecard firmware will not respond to `\r\n`
+        //       after communicating over I2C.
+        const char *transmitErr = _I2CTransmit(_I2CAddress(), (uint8_t *)"\n", 1);
+        // If we get a failure on transmitting the `\n`, it means that the
+        // Notecard isn't present.
+        if (transmitErr) {
+            NOTE_C_LOG_ERROR(ERRSTR("Failed to transmit during reset.",c_iobad));
+            return false;
+        }
+        if (!cardTurboIO) {
```
haydenroche5, line 244: Why was this added? I thought we weren't doing turbo anymore.

n_i2c.c
```diff
@@ -212,69 +212,105 @@ const char *i2cNoteTransaction(char *request, char **response, size_t timeoutMs)
...
             break;
         }
 
-    }
-
-    // Reinitialize i2c if there's no response
-    if (!notecardReady) {
-        _I2CReset(_I2CAddress());
+#ifdef ERRDBG
+        NOTE_C_LOG_ERROR(somethingFound ? ERRSTR("unrecognized data from notecard", c_iobad) : ERRSTR("notecard not responding", c_iobad));
```
haydenroche5, line 302: Should this be an error or a warning? We're going to retry (unless we're out of retries), so a warning feels more appropriate.

haydenroche5 (approved):

n_i2c.c
haydenroche5: Ah, I see it was there before. So never mind.

n_i2c.c
zfields: I completely agree and wrote it that way initially. Unfortunately, this is a ternary statement and the else is, in fact, an error. I figure it was better to promote in this case, but I'm actually not opposed to reworking the logic, so they can be different.

n_i2c.c
```diff
@@ -212,69 +212,108 @@ const char *i2cNoteTransaction(char *request, char **response, size_t timeoutMs)
...
+        // Send a newline to the module to clean out request/response processing
+        // NOTE: This MUST always be `\n` and not `\r\n`, because there are some
+        //       versions of the Notecard firmware will not respond to `\r\n`
+        //       after communicating over I2C.
+        const char *transmitErr = _I2CTransmit(_I2CAddress(), (uint8_t *)"\n", 1);
+        // If we get a failure on transmitting the `\n`, it means that the
+        // Notecard isn't present.
+        if (transmitErr) {
+            NOTE_C_LOG_ERROR(transmitErr);
+            NOTE_C_LOG_ERROR(ERRSTR("failed to reset I2C", c_bad));
```
haydenroche5, line 241: Missing `_UnlockI2C()`.

Also, just to be clear, we are intentionally not retrying here, right? I'm curious why we don't retry on `_I2CTransmit` failing, but we do retry if `_I2CReceive` fails.

haydenroche5 (changes requested):

n_i2c.c
zfields: I rewrote the function using TDD on the plane. I caught the `_UnlockI2C()` and, without thinking, I implemented the retry on transmit.

I just now ran the tests over the implementation you reviewed, and I completely agree. I have already refactored them to work just as you have suggested.

n_i2c.c
```diff
@@ -212,72 +212,119 @@ const char *i2cNoteTransaction(char *request, char **response, size_t timeoutMs)
...
+        // Determine if I2C data is available
+        // set initial state of variable to perform query
         uint16_t chunkLen = 0;
-        while (true) {
+
+        // Content flags to determine if reset conditions are met.
+        bool somethingFound = false;
+        bool nonControlCharFound = false;
+
+        // Read I2C data for at least `CARD_RESET_DRAIN_MS` continously
```
haydenroche5, line 259: continously -> continuously

n_i2c.c
```diff
@@ -212,72 +212,119 @@ const char *i2cNoteTransaction(char *request, char **response, size_t timeoutMs)
...
+        for (const size_t startMs = _GetMs() ; (_GetMs() - startMs) < CARD_RESET_DRAIN_MS ;) {
 
             // Read the next chunk of available data
-            uint32_t available;
-            uint8_t buffer[128];
+            uint32_t available = 0;
+            uint8_t buffer[ALLOC_CHUNK] = {0};
             chunkLen = (chunkLen > sizeof(buffer)) ? sizeof(buffer) : chunkLen;
             chunkLen = (chunkLen > _I2CMax()) ? _I2CMax() : chunkLen;
             delayIO();
```
haydenroche5, line 267: This can probably be removed since you added a delay of `CARD_REQUEST_I2C_CHUNK_DELAY_MS` to the end of the loop.

n_i2c.c
```diff
@@ -212,72 +212,119 @@ const char *i2cNoteTransaction(char *request, char **response, size_t timeoutMs)
...
             const char *err = _I2CReceive(_I2CAddress(), buffer, chunkLen, &available);
             if (err) {
-                break;
+                // We have received a hardware or protocol level error.
+                // Introduce delay to relieve system stress.
+                NOTE_C_LOG_ERROR(err);
+                NOTE_C_LOG_ERROR(ERRSTR("error encountered during I2C receive hook execution", c_err));
+                _DelayMs(CARD_REQUEST_I2C_SEGMENT_DELAY_MS);
+                notecardReady = false;
+                continue;
```
haydenroche5, line 276: So when we fail to receive, we try to receive again, but we don't break out of this loop and try to transmit \n again first. I'm just putting words to it to make sure that's what we intend to do.

haydenroche5 (approved):

n_i2c.c
zfields: We just want to clean the buffer. Theoretically, we are out-of-sync, and we just need to read anything back whether we fail or not.

zfields: We will always resend `\n` if we were able to receive anything other than control characters in the 500 ms.

n_i2c.c
zfields: agreed.

PR #104, opened 2023-09-25
Zak mem drain

zfields:
ensure reset will consume all available bytes,
    regardless of the amount. Previously, the 5-second
    timeout would prevent reset from consuming large
    outstanding, binary data payloads.

### DIFF
--- a/n_helpers.c
+++ b/n_helpers.c
@@ -396,6 +396,9 @@ const char * NoteBinaryStoreReceive(uint8_t *buffer, uint32_t bufLen,
 
     // Ensure transaction was successful
     if (err) {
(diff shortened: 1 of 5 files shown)


m-mcgowan:
LGTM.  Is _Reset() only ever called on error? If not, I would create a helper function _ResetAfterError() which just calls _Reset(). I feel that makes the intent clearer and gives a single point of modification should we want to add additional error handling separate from a "normal" reset.

m-mcgowan (approved):

zfields:
[Fix `note-c` memory drain behavior](https://trello.com/c/4gOwvAUL/212-fix-note-c-memory-drain-behavior)

PR #107, opened 2023-09-30
chore: standardize logging

zfields:


n_request.c
```diff
@@ -54,10 +54,10 @@ static J *errDoc(const char *errmsg)
...
 
         if (suppressShowTransactions == 0) {
-            // Since we're already allocating...
-            char * err = JConvertToJSONString(rspdoc);
-            NOTE_C_LOG_ERROR(err);
-            _Free(err);
+            _DebugWithLevel(NOTE_C_LOG_LEVEL_ERROR, "[ERROR] ");
+            _DebugWithLevel(NOTE_C_LOG_LEVEL_ERROR, "{\"err\":\"");
+            _DebugWithLevel(NOTE_C_LOG_LEVEL_ERROR, errmsg);
+            _DebugWithLevelLn(NOTE_C_LOG_LEVEL_ERROR, "\",\"src\":\"note-c\"}");
```
zfields, line 60: This is a reversion to what it used to be, but more appropriately wrapped by our logging semantics.

n_lib.h
```diff
@@ -210,9 +210,13 @@ extern const char *c_badbinerr;
 #ifdef NOTE_NODEBUG
 #define _Debug(x)
 #define _Debugln(x)
+#define _DebugWithLevel(x, y)
+#define _DebugWithLevelLn(x, y)
```
zfields, line 214: Allows the new logging to be disabled by `NOTE_NODEBUG`.

zfields:
[[note-c] Wrap "sync" logs in `NOTE_C_LOG_DEBUG`](https://trello.com/c/O2kfPI5w/215-note-c-wrap-sync-logs-in-noteclogdebug)

n_helpers.c
```diff
@@ -2341,11 +2341,10 @@ bool NoteDebugSyncStatus(int pollFrequencyMs, int maxLevel)
         J *body = JGetObject(rsp, "body");
         if (body != NULL) {
             if (maxLevel < 0 || JGetInt(body, "level") <= maxLevel) {
-                _Debug("sync: ");
-                _Debug(JGetString(body, "subsystem"));
-                _Debug(" ");
-                _Debug(JGetString(body, "text"));
-                _Debug("\n");
+                _DebugWithLevel(0, "[SYNC] ");
```
haydenroche5, line 2344: Should these 0s be `NOTE_C_LOG_LEVEL_ERROR`?

haydenroche5 (changes requested):

n_helpers.c
zfields: Yeah, but it's not an error.

I was thinking it would be better to create an alias `NOTE_C_LOG_LEVEL_ALWAYS`, but wanted to get your opinion first.

haydenroche5: I would use `NoteDebug`. Seems like there's no point in using the version of the function with log level if you want to always log it.

zfields: Yeah, that's a good point. I guess I really like the expressiveness of `NoteDebugWithLevel()`, and I hadn't even considered using the standard logging.

haydenroche5 (approved):

PR #109, opened 2023-10-06
Add a scripts/git_hooks directory with a pre-push hook.

haydenroche5:
Developers can install these hook in .git/hooks/ in their local note-c repo. The pre-push hook runs the unit tests (without valgrind, because it's slow), builds the Doxygen docs, and runs astyle. If any of those fail, the hook fails, and the push won't go through. This should save us some churn on CI builds on GitHub.


m-mcgowan (approved):
I couldn't test, since it didn't work on OSX, but if it's useful for you and Zak, I don't want to hold this up. I'll fix for OSX in a later PR...maybe. :-)

PR #111, opened 2023-10-18
Rework NoteRequestResponseJSON.

haydenroche5:
This is a much simpler function now. It does not call `NoteTransaction` under the hood, and as a result it doesn't do any JSON parsing. It does the following:

- Starts the transaction with `_TransactionStart`.
- Locks access to the Notecard.
- If the request is a command, it calls `_Transaction` with a NULL response parameter. Otherwise, it calls it with a valid pointer to hold the response.
- Unlocks access to the Notecard.
- Stops the transaction with `_TransactionStop`.
- Returns the response or NULL if there was an error or no response.

### DIFF
--- a/n_request.c
+++ b/n_request.c
@@ -283,34 +283,47 @@ J *NoteRequestResponseWithRetry(J *req, uint32_t timeoutSeconds)
  @brief Send a request to the Notecard and return the response.
 
  Unlike NoteRequestResponse, this function expects the request to be a valid
- JSON C-string, rather than a `J` object. It also returns the response as a
- dynamically allocated JSON C-string. The caller is responsible for freeing this
- string.
+ JSON C-string, rather than a `J` object. This string may be newline-terminated,
+ but it is not required. The response is returned as a dynamically allocated
+ JSON C-string. The response string is verbatim what was sent by the Notecard.
+ The terminating CR-LF will be included in the response string. The caller is
+ responsible for freeing the response string. If the request was a command (i.e.
+ it uses "cmd" instead of "req"), this function returns NULL, because the
+ Notecard does not send a response to commands.
 
  @param reqJSON A valid JSON C-string containing the request.
 
- @returns A JSON C-string with the response or NULL if there was an error
-          sending the request.
+ @returns A JSON C-string with the response or NULL if there was no response.
  */
 char *NoteRequestResponseJSON(char *reqJSON)
 {
-    // Parse the incoming JSON string
-    J *req = JParse(reqJSON);
-    if (req == NULL) {
+    size_t transactionTimeoutMs = (CARD_INTER_TRANSACTION_TIMEOUT_SEC * 1000);
+    char *rspJSON = NULL;
+
+    if (reqJSON == NULL) {
         return NULL;
     }
 
-    // Perform the transaction and free the req
-    J *rsp = NoteRequestResponse(req);
-    if (rsp == NULL) {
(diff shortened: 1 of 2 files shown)


zfields (approved):

n_request.c
```diff
@@ -283,34 +283,47 @@ J *NoteRequestResponseWithRetry(J *req, uint32_t timeoutSeconds)
...
+    if (!_TransactionStart(transactionTimeoutMs)) {
         return NULL;
     }
 
-    // Convert response back to JSON and delete it
-    char *json = JPrintUnformatted(rsp);
-    NoteDeleteResponse(rsp);
+    _LockNote();
+
+    if (strstr(reqJSON, "\"cmd\":") != NULL) {
```
m-mcgowan, line 314: I'm sure you already thought of this but `{"req":"note.add", "body": { "cmd": "open" }}`

n_request.c
```diff
@@ -283,34 +283,47 @@ J *NoteRequestResponseWithRetry(J *req, uint32_t timeoutSeconds)
  @brief Send a request to the Notecard and return the response.
 
  Unlike NoteRequestResponse, this function expects the request to be a valid
- JSON C-string, rather than a `J` object. It also returns the response as a
- dynamically allocated JSON C-string. The caller is responsible for freeing this
- string.
+ JSON C-string, rather than a `J` object. This string may be newline-terminated,
+ but it is not required. The response is returned as a dynamically allocated
+ JSON C-string. The response string is verbatim what was sent by the Notecard.
+ The terminating CR-LF will be included in the response string. The caller is
```
m-mcgowan, line 289: What's the rationale for returning the framing delimiter? Not questioning it, but would be good to know in case it comes up in future design discussions.

n_request.c
```diff
@@ -283,34 +283,47 @@ J *NoteRequestResponseWithRetry(J *req, uint32_t timeoutSeconds)
  @brief Send a request to the Notecard and return the response.
 
  Unlike NoteRequestResponse, this function expects the request to be a valid
- JSON C-string, rather than a `J` object. It also returns the response as a
- dynamically allocated JSON C-string. The caller is responsible for freeing this
- string.
+ JSON C-string, rather than a `J` object. This string may be newline-terminated,
```
m-mcgowan, line 286: Is Notecard happy not receiving the CR-LF, or is it added if not present? (I'm having trouble in github tracing `_Transaction`.). Either way, this sounds like a good candidate for HIL tests to validate that the notecard is happy with both alternatives (or as a unit test here to validate the end of frame mark is added in the code here, if that's the case.)

n_request.c
haydenroche5: This was addressed in https://github.com/blues/note-c/pull/114.

n_request.c
haydenroche5: I think the idea is to return exactly what's received from the Notecard, which includes the \n. But I don't think it's critical, personally.

n_request.c
haydenroche5: The Notecard is fine with either \r\n or \n. It's weird, but for serial, we always terminate the request to the Notecard with \r\n. For I2C, it's always \n. @zfields will remember better why Ray wanted it this way; I don't know/remember the details.

zfields: The Windows 7 USB driver was not guaranteed to flush streams that did not end `\r\n`, which is why it exists on serial communication. The Notecard is perfectly content with a solitary `\n` to punctuate a Note.

PR #113, opened 2023-10-20
change json contract

rayozzie:


zfields:
Used as input for #114

PR #114, opened 2023-10-20
Improve c-string API

zfields:
- standardize newline termination behavior
- accept chain "cmd" request
- handle string literals


zfields (commented):
looks good to me.

PR #116, opened 2023-10-23
Add `void` to functions taking no params.

haydenroche5:
See https://softwareengineering.stackexchange.com/a/286494.

### DIFF
--- a/n_cjson.c
+++ b/n_cjson.c
@@ -164,7 +164,7 @@ N_CJSON_PUBLIC(void) JFree(void *p)
 }
 
 /* Internal constructor. */
-static J *JNew_Item()
+static J *JNew_Item(void)
 {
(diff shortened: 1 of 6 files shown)


zfields:
This is obviously forces the API to be used as we intended, but does it merit bumping the version of the API, since the API contract has actually changed?

zfields:
> This is obviously forces the API to be used as we intended, but does it merit bumping the version of the API, since the API contract has actually changed?

I'm not sure it does. It's possible that some code would no longer complie, but it would not need to be materially changed to support the validated calling convention.

zfields (approved):

haydenroche5:
@zfields FWIW all these functions were all declared as `(void)`. They were just missing it in their definitions. I think because of this nothing has actually changed.

PR #126, opened 2023-12-11
Limit scope of libc usage in note-c.

haydenroche5:
See commit messages for more details.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,7 @@ option(NOTE_C_BUILD_TESTS "Build tests." OFF)
 option(NOTE_C_COVERAGE "Compile for test NOTE_C_COVERAGE reporting." OFF)
 option(NOTE_C_MEM_CHECK "Run tests with Valgrind." OFF)
 option(NOTE_C_BUILD_DOCS "Build docs." OFF)
+option(NOTE_C_NO_LIBC "Build the library without linking against libc, generating errors for any undefined symbols." OFF)
 
 set(NOTE_C_SRC_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 add_library(note_c SHARED)
@@ -55,6 +56,16 @@ target_include_directories(
     note_c
     PUBLIC ${NOTE_C_SRC_DIR}
 )
+
+if(NOTE_C_NO_LIBC)
+    target_link_options(
+        note_c
+        PRIVATE
+            -nostdlib
+            -nodefaultlibs
+            LINKER:--no-undefined
+    )
+endif()
 
 if(NOTE_C_BUILD_TESTS)
     # Including CTest here rather than in test/CMakeLists.txt allows us to run
--- a/n_cjson.c
+++ b/n_cjson.c
@@ -417,7 +417,7 @@ static Jbool print_number(const J * const item, printbuffer * const output_buffe
     /* This checks for NaN and Infinity */
     if ((d * 0) != 0) {
         char *nbuf = (char *) number_buffer;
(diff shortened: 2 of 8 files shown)


n_cjson.c
```diff
@@ -115,6 +121,21 @@ N_CJSON_PUBLIC(const char*) JVersion(void)
     return NOTE_C_STRINGIZE(N_CJSON_VERSION_MAJOR) "." NOTE_C_STRINGIZE(N_CJSON_VERSION_MINOR) "." NOTE_C_STRINGIZE(N_CJSON_VERSION_PATCH);
 }
 
+NOTE_C_STATIC char Jtolower(char c)
+{
+    // Less than A -> not a valid letter.
+    // Greater than z -> not a valid letter.
+    // Between a and Z -> not a valid letter.
+    // Between a and z (inclusive) -> already lower.
+    if (c < 'A' || c > 'z' || (c < 'a' && c > 'Z') || (c >= 'a' && c <= 'z')) {
```
zfields, line 130: This seems overly complicated.

It becomes more obvious if you swap in numbers.

Return the letter if:
x < 65 OR x > 122 OR (x < 97 AND x > 90) OR (x >= 97 AND x <= 122)

-128 >= x < 65,
x > 122,
90 < x < 97,
97 <= x <= 122

-128 <------ 65 ______ 91 --97--122--> 127

```suggestion
    if (c < 'A' || c > 'Z') {
```

scripts/check_libc_dependencies.sh
```diff
@@ -0,0 +1,96 @@
...
+PERMITTED_FNS=(
+    # The mem* functions are ok.
+    "memchr"
+    "memcmp"
+    "memcpy"
+    "memmove"
+    "memset"
+    # These string functions are ok.
+    "strchr"
+    "strcmp"
```
zfields, line 31: This seems like a good opportunity to not allow these unsafe functions.

CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ option(NOTE_C_BUILD_TESTS "Build tests." OFF)
 option(NOTE_C_COVERAGE "Compile for test NOTE_C_COVERAGE reporting." OFF)
 option(NOTE_C_MEM_CHECK "Run tests with Valgrind." OFF)
 option(NOTE_C_BUILD_DOCS "Build docs." OFF)
+option(NOTE_C_NO_LIBC "Build the library without linking against libc, generating errors for any undefined symbols." OFF)
```
zfields, line 21: Is this supposed to be off, then gets switched on for specific test builds?

scripts/check_libc_dependencies.sh
zfields: FWIW, I don't think this should prevent this commit from happening, but we should add it to the backlog - especially now that there is a mechanism for protecting against this.

CMakeLists.txt
haydenroche5: Yes.

n_cjson.c
haydenroche5: Nah I think your change works.

scripts/check_libc_dependencies.sh
haydenroche5: I'd have to re-immerse myself in the discourse on this age-old argument, but my rough understanding is that you're not that much better off using strncmp over strcmp: https://stackoverflow.com/questions/24353504/whats-wrong-with-strcmp

If your string isn't NULL-terminated, all bets are off.

zfields (approved):

PR #131, opened 2024-01-11
Fix timeout data type bugs.

haydenroche5:
This commit resolves a couple problems.

First, it changes the return type of NoteGetMs. NoteGetMs is simply a wrapper around the user-supplied millisecond timer hook. The prototype for this hook specifies that it should return a uint32_t yet NoteGetMs's return type is long unsigned int. This commit changes the return type of NoteGetMs to uint32_t.

Second, there are many places in the code where we are comparing and assigning millisecond timer values (i.e. gotten via NoteGetMs) to size_t variables. This is a bug. On some platforms (e.g. the Arduino Uno), size_t is 16 bits, so assigning one of these 32-bit millisecond timer values to a 16-bit variable can lead to undefined behavior when the timer value is larger than what can fit into 16 bits. This commit fixes all these bugs by using uint32_t instead of size_t.

### DIFF
--- a/n_hooks.c
+++ b/n_hooks.c
@@ -174,8 +174,8 @@ i2cReceiveFn hookI2CReceive = NULL;
 
 // Internal hooks
 typedef bool (*nNoteResetFn) (void);
-typedef const char * (*nTransactionFn) (const char *, size_t, char **, size_t);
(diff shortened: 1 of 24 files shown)


haydenroche5:
I'll have to dig into why the `card.binary` tests failed later. It looks like some part of my home HIL infrastructure is down.

zfields (approved):

PR #132, opened 2024-01-18
Improve cJSON number handling.

haydenroche5:
- Get rid of CJSON_NO_CLIB and its associated code paths, per Ray's request.
- Change the type of the J object's valueint member. It should be int32_t if NOTE_LOWMEM and int64_t otherwise. Capture this in a new #define for the type, JINTEGER.
- When parsing a JSON number from a JSON string, if the number is within the bounds of the JINTEGER type, convert it to an integer using JAtoI rather than simply casting valuenumber to JINTEGER, which can lose precision (e.g. for a Unix timestamp). If the number is outside the bounds of a JINTEGER, saturate, which is the same behavior as before this commit.
- When printing a JSON number, print it as an integer if possible. Otherwise, print it as floating point.
- Fix an issue where JItoA would fail to convert the minimum long int value (i.e. LONG_MIN) to a string.

### DIFF
--- a/n_atof.c
+++ b/n_atof.c
@@ -257,7 +257,7 @@ char **endPtr;              /* If non-NULL, store terminating character's
         case 5:
             p10 = 1.0e32;
             break;
-#ifndef NOTE_FLOAT
+#ifndef NOTE_LOWMEM
         case 6:
             p10 = 1.0e64;
             break;
--- a/n_cjson.c
+++ b/n_cjson.c
@@ -63,9 +63,6 @@
 
 // For Note, disable dependencies
 #undef ENABLE_LOCALES
-#ifndef CJSON_NO_CLIB
-#define CJSON_NO_CLIB      1       // Use tiny but non-robust versions of conversions
-#endif
 
 #include "n_lib.h"
 
@@ -287,25 +284,19 @@ static Jbool parse_number(J * const item, parse_buffer * const input_buffer)
 loop_end:
     number_c_string[i] = '\0';
 
-    /* some platforms may not have locale support */
-#if !CJSON_NO_CLIB
-    number = strtod((const char*)number_c_string, (char**)&after_end);
-#else
     number = JAtoN((const char*)number_c_string, (char**)&after_end);
-#endif
     if (number_c_string == after_end) {
         return false; /* parse_error */
     }
-
     item->valuenumber = number;
 
-    /* use saturation in case of overflow */
-    if (number >= LONG_MAX) {
-        item->valueint = LONG_MAX;
-    } else if (number <= LONG_MIN) {
-        item->valueint = LONG_MIN;
+    // Saturate valueint in the case of overflow.
+    if (number >= JINTEGER_MAX) {
+        item->valueint = JINTEGER_MAX;
+    } else if (number <= JINTEGER_MIN) {
+        item->valueint = JINTEGER_MIN;
     } else {
-        item->valueint = (long int)number;
+        item->valueint = JAtoI((const char*)number_c_string);
     }
 
     item->type = JNumber;
@@ -320,12 +311,14 @@ N_CJSON_PUBLIC(JNUMBER) JSetNumberHelper(J *object, JNUMBER number)
     if (object == NULL) {
         return number;
     }
-    if (number >= LONG_MAX) {
-        object->valueint = LONG_MAX;
-    } else if (number <= LONG_MIN) {
-        object->valueint = LONG_MIN;
+
+    // Saturate valueint in the case of overflow.
+    if (number >= JINTEGER_MAX) {
+        object->valueint = JINTEGER_MAX;
+    } else if (number <= JINTEGER_MIN) {
+        object->valueint = JINTEGER_MIN;
     } else {
-        object->valueint = (long int)number;
+        object->valueint = (JINTEGER)number;
     }
 
     return object->valuenumber = number;
@@ -421,7 +414,8 @@ static Jbool print_number(const J * const item, printbuffer * const output_buffe
     }
 
     unsigned char *output_pointer = NULL;
-    JNUMBER d = item->valuenumber;
+    JNUMBER vnum = item->valuenumber;
+    JINTEGER vint = item->valueint;
     int length = 0;
     size_t i = 0;
     unsigned char number_buffer[JNTOA_MAX]; /* temporary buffer to print the number into */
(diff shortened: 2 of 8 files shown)


n_cjson.c
```diff
@@ -61,12 +61,6 @@
 #include <limits.h>
 #include <ctype.h>
 
-// For Note, disable dependencies
-#undef ENABLE_LOCALES
```
haydenroche5, line 65: Didn't mean to delete this line. I'll add it back.

haydenroche5:
Added unit tests.

n_cjson.c
```diff
@@ -287,25 +284,19 @@ static Jbool parse_number(J * const item, parse_buffer * const input_buffer)
...
-    } else if (number <= LONG_MIN) {
-        item->valueint = LONG_MIN;
+    // Saturate valueint in the case of overflow.
+    if (number >= JINTEGER_MAX) {
+        item->valueint = JINTEGER_MAX;
+    } else if (number <= JINTEGER_MIN) {
+        item->valueint = JINTEGER_MIN;
     } else {
-        item->valueint = (long int)number;
+        item->valueint = JAtoI((const char*)number_c_string);
```
sdt99, line 299: JAtoI needs to be modified to return type JINTEGER
`long int JAtoI(const char *string)`

n_cjson_helpers.c
```diff
@@ -431,15 +431,18 @@ const char *JGetItemName(const J * item)
 void JItoA(long int n, char *s)
 {
     char c;
-    long int i, j, sign;
-    if ((sign = n) < 0) {
-        n = -n;
+    // Conversion to unsigned is required to handle the case where n is
+    // LONG_MIN.
```
zfields, line 435: This comment is not obvious. Why is this the case?

n_cjson_helpers.c
```diff
@@ -431,15 +431,18 @@ const char *JGetItemName(const J * item)
...
+    unsigned long int unsignedN = n;
+    long int i, j;
+    if (n < 0) {
+        unsignedN = -unsignedN;
```
zfields, line 439: The data type of `unisignedN` is `unsigned long int`, so what does making it negative do?

note.h
```diff
@@ -25,28 +25,43 @@
...
+#define NOTE_LOWMEM
 #endif
 #else
 #error What are floating point exponent length symbols for this compiler?
 #endif
 
-// If using a short float, we must be on a VERY small MCU.  In this case, define additional
-// symbols that will save quite a bit of memory in the runtime image.
-#ifdef NOTE_FLOAT
+#ifndef JNUMBER
```
zfields, line 42: The placement of this `#if` :point_up:  is not obvious. Why is it not below `#ifdef NOTE_LOWMEM`?

I think it would make it easier to understand the subsequent `#if` logic.

n_atof.c
```diff
@@ -257,7 +257,7 @@ char **endPtr;              /* If non-NULL, store terminating character's
         case 5:
             p10 = 1.0e32;
             break;
-#ifndef NOTE_FLOAT
+#ifndef NOTE_LOWMEM
```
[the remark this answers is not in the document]
haydenroche5: I would like to yeah. In theory there could be users depending on this macro. Should we change the define to `NOTE_C_LOWMEM`, but also define `NOTE_LOWMEM` if `NOTE_C_LOWMEM` is defined for backwards compatibility?

The same goes for `NOTE_FLOAT`. I opted to eliminate it entirely, but there could be someone out there using it in their code.

n_cjson.c
```diff
@@ -287,25 +284,19 @@ static Jbool parse_number(J * const item, parse_buffer * const input_buffer)
...
-#if !CJSON_NO_CLIB
-    number = strtod((const char*)number_c_string, (char**)&after_end);
-#else
     number = JAtoN((const char*)number_c_string, (char**)&after_end);
-#endif
     if (number_c_string == after_end) {
         return false; /* parse_error */
     }
-
     item->valuenumber = number;
```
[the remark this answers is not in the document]
haydenroche5: This is storing the `JNUMBER` version of the string number, obtained with `JAtoN`. We separately convert the string number to a `JINTEGER` with `JAtoI`.

n_cjson.c
haydenroche5: This gets complicated. Strap in!

When we're printing (instead of parsing), we do this:

```
if (vnum != (JNUMBER)vint) {
    JNtoA(vnum, nbuf, -1);
} else {
    JItoA(vint, nbuf);
}
```

`vnum == (JNUMBER)vint` for all integers within a certain range, depending on if JNUMBER is single or double precision. Past a certain range, each step for the floating point number (`vnum`) is greater than 1 and so you start to not be able to represent every integer exactly. See the "Precision limitations on integer values" section here: https://en.wikipedia.org/wiki/Single-precision_floating-point_format

So for certain large integers we simply can't store them exactly.

With that out of the way, for your actual question

> Do we have any way to know if the value is actually JINTEGER_MAX or an overflow?

No. Let's say we're in the `NOTE_LOWMEM` case so `JINTEGER` is `int32_t` and `JNUMBER` is `float` (single precision). Let's say we parse a JSON string with a field that's `JINTEGER_MAX + 1`. That's 2147483648. This is 2^31. Following this rule from that section of the Wikipedia page

> Integers between 2^(n) and 2^(n+1) round to a multiple of 2^(n-23).

we know that the `JNUMBER` stored when parsing will be 2^(31-23), which is 256. Indeed it gets rounded to 2147483650, which is the nearest multiple of 256. So `valueint` is 2147483647 and `valuenum` is 2147483650. Both are inaccurate, and storing a bit to indicate there was overflow doesn't help us, because neither type can precisely represent the original number. When it comes time to convert the `J` object back to a string again, the `valueint` version is what gets printed, because `vnum == (JNUMBER)vint`. If you keep increasing the value, you'll get to a point where 2147483650 is no longer the nearest multiple of 256, and we'll print `vnum` instead.

I've tried to address all this edge case stuff in the unit tests I added.

n_cjson_helpers.c
haydenroche5: If n is `JINTEGER_MIN`, when we negate it with the unary minus operator, it overflows and the result is undefined behavior (I'm pretty sure). If I keep the original code that was here and pass in `LONG_MIN`, this

```
n = -n;
```

Just keeps n as `LONG_MIN`. Then you end up doing `n % 10` on a negative number and the whole algorithm breaks down. If we cast n to unsigned, then the rules for negation are different. See: https://stackoverflow.com/questions/8026694/c-unary-minus-operator-behavior-with-unsigned-operands

> The negative of an unsigned quantity is computed by subtracting its value from 2^n, where n is the number of bits in the promoted operand.

So `unsignedN = -unsignedN` when `JINTEGER` is 32 bits and n is `JINTEGER_MIN` yields 2^32-2147483648, which is 2147483648. Then the '% 10' algorithm works as intended.

I'll explain all this better in an improved code comment.

n_cjson_helpers.c
haydenroche5: See above.

note.h
haydenroche5: The idea here was to allow the user to define `JNUMBER` themselves, if desired. Then none of our  preprocessor logic here applies. Their definition takes precedence. So I read it as "if the user hasn't specified what `JNUMBER` should be already, apply these rules."

n_cjson.c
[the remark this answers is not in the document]
zfields: Yeah, but it is definitely overwritten. There is no circumstance where it is left as is.

n_atof.c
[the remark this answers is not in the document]
zfields: You're right. I think we should change it to `NOTE_C_LOWMEM`, and also declare `NOTE_LOWMEM`. There should probably be a deprecation message before it is `#define`'d, in case it was declared by the build system.

note.h
zfields: So if the user provides their own `JNUMBER`, then we don't need to consider if `NOTE_LOWMEM` is defined?

n_cjson.c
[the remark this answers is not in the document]
zfields: I didn't notice that it was `valuenumber` as opposed to `valueint`

n_cjson.c
zfields: It seems like it is possible to tell the difference, because you can check if `valuenumber` != `valueint`.

n_cjson_helpers.c
zfields: That explanation makes more sense when the data type is signed. I don't see how this works with an unsigned data type.

zfields: `JINTEGER_MIN` should be `0x80000000`, right?

If you make that unsigned, then you have a valid positive number.
`unsigned long int unsignedN = n;`

Then if you make it negative with `-`. Now, you have a temporary 64-bit negative number.
`0xFFFFFFFF80000000`, which you then assign to a 32-bit unsigned number, so you are right back to a valid 32-bit unsigned number `0x80000000`.

zfields: I understand that modulo will work better on unsigned values, but I don't quite get the point of setting `unsignedN` to `-unsignedN`.

PR #133, opened 2024-01-22
Improve JSON number handling (continued).

haydenroche5:
Make more fixes to number handling and NOTE_C_LOW_MEM code.

- Remove #ifdef ERRDBG guard in i2cNoteReset. This was causing us to not call
_I2CReset when we should have been.
- Make JTIME a typedef for JUINTEGER.
- Fix run_unit_tests.sh script so that it actually runs the tests with
NOTE_C_LOW_MEM defined.
- Use JAddIntToObject to add Unix timestamps to J objects in unit tests. This is
necessary to avoid loss of precision when using JAddNumberToObject.

---

Improve JSON number handling (continued).

- Change NOTE_LOWMEM to NOTE_C_LOW_MEM. If NOTE_C_LOW_MEM is defined, the old
macro NOTE_LOWMEM is automatically defined. Same for NOTE_FLOAT, which was also
removed.
- Change JItoA, JAtoI, NoteSetEnvDefaultInt, and NoteGetEnvInt to use JINTEGER
instead of long int.
- Make JINTEGER a typedef for int64_t. Don't bother trying to save a few bytes
in the NOTE_C_LOW_MEM case.
- Make JNUMBER a typedef for double. On low memory platforms, double is often
the same as float anyway (e.g. AVR).
- Improve comment explaining the JINTEGER_MIN case in JItoA.
- Simplify JSON_number_handling_test.cpp.

### DIFF
--- a/note.h
+++ b/note.h
@@ -39,12 +39,12 @@
 #error What are floating point exponent length symbols for this compiler?
 #endif
 
-// NOTE_LOWMEM is the old name of NOTE_C_LOW_MEM. If NOTE_C_LOW_MEM is defined,
-// we define NOTE_LOWMEM as well, for backwards compatibility. NOTE_FLOAT is
+// NOTE_LOWMEM is the old name of NOTE_C_LOW_MEM. If NOTE_LOWMEM is defined,
+// we define NOTE_C_LOW_MEM as well, for backwards compatibility. NOTE_FLOAT is
 // also no longer used internally, but used to be defined when NOTE_LOWMEM was
 // defined. It's also preserved here for backwards compatibility.
-#ifdef NOTE_C_LOW_MEM
-#define NOTE_LOWMEM
+#ifdef NOTE_LOWMEM
+#define NOTE_C_LOW_MEM
 #define NOTE_FLOAT
 #endif
 
@@ -66,7 +66,7 @@ typedef uint64_t JUINTEGER;
 // UNIX Epoch time (also known as POSIX time) is the  number of seconds that have elapsed since
 // 00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC).  In this project, it always
 // originates from the Notecard, which synchronizes the time from both the cell network and GPS.
-typedef unsigned long int JTIME;
+typedef JUINTEGER JTIME;
 
 // C-callable functions
 #ifdef __cplusplus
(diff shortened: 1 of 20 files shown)


sdt99 (approved):
Looks good to me.

sdt99 (approved):
I ported this into NOTE-LORA and verified that it fixed the handling of large int64 values in e-e tests.
Approved - subject to Zak's feedback.

haydenroche5:
@zfields Sorry for cramming a lot of seemingly different things into this PR, but a lot of it is actually related, and it's hard to logically separate out into multiple PRs.

note.h
```diff
@@ -26,50 +26,47 @@
...
-#define JINTEGER int64_t
-#define JINTEGER_MIN INT64_MIN
-#define JINTEGER_MAX INT64_MAX
-#endif
+// NOTE_LOWMEM is the old name of NOTE_C_LOW_MEM. If NOTE_C_LOW_MEM is defined,
+// we define NOTE_LOWMEM as well, for backwards compatibility. NOTE_FLOAT is
+// also no longer used internally, but used to be defined when NOTE_LOWMEM was
+// defined. It's also preserved here for backwards compatibility.
+#ifdef NOTE_C_LOW_MEM
+#define NOTE_LOWMEM
```
zfields, line 47: This is backwards.

There is no longer any code gated by `NOTE_LOWMEM`. If `NOTE_LOWMEM` is defined by the build system, then we should define `NOTE_C_LOW_MEM` to reactive all the gates.
```suggestion
#ifdef NOTE_LOWMEM
#define NOTE_C_LOW_MEM
```

zfields (commented):
I understand the need for this, but I am concerned we are becoming more tightly coupled with the JSON implementation instead of becoming more abstracted.

sdt99:
The updates make sense to me.
I think I understand the motivation behind making JINTEGER always an int64 type - to allow epoch times to be stored in a JINTEGER without worries about sign problems.

haydenroche5:
> The updates make sense to me. I think I understand the motivation behind making JINTEGER always an int64 type - to allow epoch times to be stored in a JINTEGER without worries about sign problems.

Yeah,

note.h
zfields: Once this is addressed :point_up_2:, then I think we are good to merge.

PR #136, opened 2024-01-25
Make more JSON number handling improvements.

haydenroche5:
- Incorporate Ray's improvements to n_ftoa.c and n_atof.c that allow us to support larger exponents.
- Add JAtoN and JNtoA unit tests to test these improvements.
- If NOTE_C_LOW_MEM is defined, define JNUMBER to float. On truly low memory platforms, this likely does nothing, because those platforms often map double and float to the same underlying type (which is single-precision). However, when running the unit tests on a CPU with double-precision floating point support, this allows us to deliberately run the unit tests with JNUMBER set to single-precision. Doing this broke a couple test cases in JSON_number_handling_test.cpp, which I've fixed here.

### DIFF
--- a/note.h
+++ b/note.h
@@ -55,7 +55,11 @@
 #define ERRDBG
 #endif
 
+#ifdef NOTE_C_LOW_MEM
+typedef float JNUMBER;
+#else
 typedef double JNUMBER;
+#endif
 
 typedef int64_t JINTEGER;
 #define JINTEGER_MIN INT64_MIN
--- a/note.h
+++ b/note.h
@@ -55,7 +55,7 @@
 #define ERRDBG
 #endif
 
-#ifdef NOTE_C_LOW_MEM
+#ifdef NOTE_C_TEST_SINGLE_PRECISION
 typedef float JNUMBER;
 #else
 typedef double JNUMBER;
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,7 +139,7 @@ jobs:
 
       - name: Run tests with NOTE_LOWMEM defined
         run: |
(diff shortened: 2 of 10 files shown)


haydenroche5:
Context: https://github.com/blues/notecard/pull/788

note.h
```diff
@@ -55,7 +55,11 @@
 #define ERRDBG
 #endif
 
+#ifdef NOTE_C_LOW_MEM
```
zfields, line 58: How much value is there in making this distinction? Do we know how much memory we are saving by switching down to `float` from `double`?

This feels very similar to the `JINTEGER` discussion. My gut instinct is to only support `double`, unless we have a compelling reason to do otherwise.

haydenroche5: So the thinking is that most small MCUs define double and float to the same thing: a single-precision (32-bit) float. That's what drives the logic on lines 28-40. The primary motivator here is testing. If I'm running the unit tests on my 64-bit Linux machine, I can define NOTE_C_LOW_MEM and run the tests using 32-bit floats instead of doubles, without the need for an actual platform like an Uno where there's ONLY 32-bit floats (no doubles).

zfields: In that case, it seems as if it should be under a test specific define, instead of the `NOTE_C_LOW_MEM` define.

haydenroche5: Yeah I got there eventually, too, after giving it more thought. I'll tweak this PR accordingly.

rayozzie: FWIW I would caution against assuming that UNO's broken compiler is representative of "most" MCU's.  I don't know who "most" really is.  

For example, I could argue that whatever compilers are used for native PIC or MSP430 probably define the majority.   

Net-net, having specific defines would be the best, with default internal behaviors being the "most correct, least gotchas" of 64-bit int64_t and 64-bit double.

zfields (approved):

PR #140, opened 2024-01-31
Remove ERRDBG.

haydenroche5:
ERRDBG is a macro that's used to conditionally compile in various error logs. It's not used consistently, and we also already support log levels and NOTE_NODEBUG, which are more than enough to control the desired level of logging at compile-time.

### DIFF
--- a/n_i2c.c
+++ b/n_i2c.c
@@ -58,18 +58,14 @@ NOTE_C_STATIC const char * i2cNoteQueryLength(uint32_t * available,
         // Send a dummy I2C transaction to prime the Notecard
         const char *err = _I2CReceive(_I2CAddress(), &dummy_buffer, 0, available);
         if (err) {
-#ifdef ERRDBG
             NOTE_C_LOG_ERROR(err);
-#endif
             return err;
         }
 
         // If we've timed out, return an error
         if (timeoutMs && _GetMs() - startMs >= timeoutMs) {
             const char *err = ERRSTR("timeout: no response from Notecard {io}", c_iotimeout);
-#ifdef ERRDBG
             NOTE_C_LOG_ERROR(err);
-#endif
             return err;
         }
     }
@@ -119,9 +115,7 @@ const char *i2cNoteTransaction(const char *request, size_t reqLen, char **respon
     uint32_t available = 0;
     err = i2cNoteQueryLength(&available, timeoutMs);
     if (err) {
-#ifdef ERRDBG
         NOTE_C_LOG_ERROR(ERRSTR("failed to query Notecard", c_err));
-#endif
         _UnlockI2C();
         return err;
     }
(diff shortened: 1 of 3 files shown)


n_serial.c
```diff
@@ -93,9 +89,7 @@ const char *serialNoteTransaction(const char *request, size_t reqLen, char **res
         const char *err = serialChunkedReceive((uint8_t *)(jsonbuf + jsonbufLen), &jsonbufAvailLen, true, (CARD_INTRA_TRANSACTION_TIMEOUT_SEC * 1000), &available);
         if (err) {
             _Free(jsonbuf);
-#ifdef ERRDBG
-            NOTE_C_LOG_ERROR(ERRSTR("error occured during receive", c_iobad));
-#endif
+            NOTE_C_LOG_DEBUG(ERRSTR("error occured during receive", c_iobad));
```
haydenroche5, line 92: @zfields Ping me if you want me to explain why the change to NOTE_C_LOG_DEBUG from NOTE_C_LOG_ERROR. I did that in a few places. Easier to explain verbally.

n_i2c.c
```diff
@@ -58,18 +58,13 @@ NOTE_C_STATIC const char * i2cNoteQueryLength(uint32_t * available,
...
             NOTE_C_LOG_ERROR(err);
-#endif
             return err;
         }
 
         // If we've timed out, return an error
         if (timeoutMs && _GetMs() - startMs >= timeoutMs) {
             const char *err = ERRSTR("timeout: no response from Notecard {io}", c_iotimeout);
-#ifdef ERRDBG
-            NOTE_C_LOG_ERROR(err);
```
zfields, line 71: Did you intend to remove the actual log as well?

n_i2c.c
```diff
@@ -414,11 +399,9 @@ const char *i2cChunkedReceive(uint8_t *buffer, uint32_t *size, bool delay, uint3
         // Exit on timeout
         if (timeoutMs && (_GetMs() - startMs >= timeoutMs)) {
             *size = received;
-#ifdef ERRDBG
             if (received) {
-                NOTE_C_LOG_ERROR(ERRSTR("received only partial reply before timeout", c_iobad));
+                NOTE_C_LOG_DEBUG(ERRSTR("received only partial reply before timeout", c_iobad));
```
zfields, line 403: Why the downgrade from `ERROR` to `DEBUG`?

IIRC, this means the response is invalid, and the transaction must be resubmitted.

I understand downgrading if it is in a retry loop, so long as it is eventually reported as an error once the retries have been exhausted.

n_serial.c
zfields: Ha! Okay, sure. You can disregard my previous note, I was working my way down before I saw this message.

n_serial.c
```diff
@@ -190,11 +182,7 @@ bool serialNoteReset(void)
             break;
         }
 
-#ifdef ERRDBG
-        NOTE_C_LOG_ERROR(somethingFound ? ERRSTR("unrecognized data from notecard", c_iobad) : ERRSTR("notecard not responding", c_iobad));
```
zfields, line 194: Seems like we should keep this message instead of the other one.

n_serial.c
```diff
@@ -190,11 +182,7 @@ bool serialNoteReset(void)
             break;
         }
 
-#ifdef ERRDBG
-        NOTE_C_LOG_ERROR(somethingFound ? ERRSTR("unrecognized data from notecard", c_iobad) : ERRSTR("notecard not responding", c_iobad));
-#else
         NOTE_C_LOG_ERROR(ERRSTR("notecard not responding", c_iobad));
```
zfields, line 185: Swap this one out for the preceding message.

zfields (approved):

zfields (approved):

PR #142, opened 2024-03-13
fix: when sending unsupported requests to a notecard, don't wastefully retry

rayozzie:
Notecards of all types, when they get an unrecognized request, return both {io} and {not-supported} in their err field.

This is because the notecard has no idea whether the request type is corrupted because of bad I/O, or because it is not supported.

As such, when a host app innocently does a card.wireless expecting to get a {not-supported} error, it takes ten seconds or more to get the failure because it is retrying it over and over because of the {io}.

Although non-zero, the reality is that there is lower chance of it being an {io} error if in fact {not-supported} is found.

As such, I've opted to not do retries if that specific error comes back.


rayozzie:
i already made this fix to note-go and i would strongly recommend that you make it to note-python

zfields (approved):

PR #143, opened 2024-03-14
Improve run_tunnelmole.sh

haydenroche5:
- Make script easier to read. None of this business with logical operators where an "if" would be simpler.
- Improve logging messages.
- Push all logs to stderr, where they belong.

By the way, I think my Orange Pi just needed to be rebooted for things to start working for `card.binary` HIL. It was hard to tell what was going on with the tunnelmole bit, though, so I went ahead and improved the script that runs that here.

### DIFF
--- a/.github/workflows/notecard-binary-tests.yml
+++ b/.github/workflows/notecard-binary-tests.yml
@@ -94,7 +94,7 @@ jobs:
         uses: JarvusInnovations/background-action@v1
         with:
           run: |
-            bash ./scripts/run_md5srv.sh
+            ./scripts/run_md5srv.sh
           wait-on:
             # just a dummy wait-on since this is required.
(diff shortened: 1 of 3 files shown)


test/hitl/card.binary/lib/notecard_binary/NotecardComms.h
haydenroche5, line 1: astyle...

.github/workflows/notecard-binary-tests.yml
```diff
@@ -94,7 +94,7 @@ jobs:
         uses: JarvusInnovations/background-action@v1
         with:
           run: |
-            bash ./scripts/run_md5srv.sh
```
haydenroche5, line 97: No need to invoke `bash`. Just makes things more complicated.

zfields (approved):

PR #149, opened 2024-05-17
TS-513: Add JPrintPreallocated to note-c.

haydenroche5:
We added this function to the note-c copy used in the NOTE-LORA firmware with NF-473.

We aim to keep the note-c used in the firmware the same as the upstream note-c we distribute as an SDK, so we need to bring this function over.


haydenroche5:
Ugh, I really need to get the card.binary HIL tests stable...

zfields (approved):

PR #153, opened 2024-06-27
TS-534: Fix card.binary HIL testing.

haydenroche5:
- Use a single PlatformIO command to build, upload, and test. Previously, we
we did this in a few steps because of a problem where the Swan's serial port
wasn't ready after uploading and so the test step would fail because it couldn't
find the serial port. I was able to reduce this to one step by using the
approach described here: https://github.com/platformio/platformio-core/issues/3742
- Fix a Bash typo in the step that cleaned up the Notehub proxy route.
- I reinstalled tunnelmole using npm on the Orange Pi, and it's working again.
I really don't know what was wrong here. In order for run_tmole.sh to find the
tmole executable, I had to add a couple lines to the start of the script so that
npm packages are on PATH.
- Upgrade the note-arduino version to latest.


haydenroche5:
Ugh, don't review this yet. There are some kinks that need to be worked out.

zfields:
Doesn't imply all tests should pass now?

zfields (approved):

PR #158, opened 2024-10-08
TS-534: Get card.binary HIL tests working again.

haydenroche5:
- Use Notestation (hil_lab) infrastructure.
- Use ngrok instead of tunnelmole. The latter has proven to be unreliable.
- Fix the md5srv-tests, which were failing due to some issue with the bats action we were using. It's now deprecated, and I moved us to a different action that the maintainer points to as the successor.


zfields (approved):

test/hitl/scripts/md5srv.py
haydenroche5, line 1: FYI, these are all just whitespace changes. I have an auto-formatter thing that runs on save when I save a .py file. Can revert if desired.

PR #159, opened 2024-11-09
chore: rename non-public API

zfields:


haydenroche5 (approved):
Looks good to me. I didn't bother reviewing all the test files. I assume if the tests are building and passing, you did everything right. :)

haydenroche5:
I need to debug the HIL card.binary tests (again!). Cancelled the job here that was hung and disabled the workflow, for now.

PR #160, opened 2024-11-18
NF-613: Fix card.binary HIL tests.

haydenroche5:
- Hardcode Notestation to los-angeles-notestation-1. This one's known to be set up for running these tests. I have work in progress to make it so this doesn't need to be hardcoded (i.e. the Notestation client can discover which stations can run these tests), but it's not finished, yet.
- Make sure to kill the Notestation client in the GitHub Actions workflow when done with the Notestation.
- Move the `file .pio/build/debug/firmware.elf` gdb command earlier in the custom upload command for PlatformIO. This just gets rid of some warnings I was seeing in the log. No functional difference.
- Change test_port to /tmp/ns_mcu_usb from /tmp/mcu_usb. The former is the new default path for the MCU USB device with the latest Notestation client.

### DIFF
--- a/.github/workflows/notecard-binary-tests.yml
+++ b/.github/workflows/notecard-binary-tests.yml
@@ -186,16 +186,17 @@ jobs:
           cd hil_lab/
           pip install -r requirements.txt
 
-          # Reserve a Notestation with the label "note_c_card_binary_test",
-          # which means it has everything needed to run the note-c card.binary
-          # tests.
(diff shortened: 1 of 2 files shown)


haydenroche5:
@zfields I needed to fix some Notestation infrastructural stuff, and now with that done + the changes in this PR, we're green: https://github.com/blues/note-c/actions/runs/12551283976

PR #162, opened 2025-01-22
fix: {io} logs

zfields:
All ERROR messages were being reported as {io} errors

### DIFF
--- a/test/src/_crcError_test.cpp
+++ b/test/src/_crcError_test.cpp
@@ -17,72 +17,175 @@
 
 #include "test_static.h"
 
+extern bool notecardFirmwareSupportsCrc;
+
 namespace
 {
 
-SCENARIO("_crcError")
-{
-    NoteSetFnDefault(malloc, free, NULL, NULL);
+    SCENARIO("_crcError")
+    {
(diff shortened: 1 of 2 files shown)


test/src/_crcError_test.cpp
```diff
@@ -17,72 +17,175 @@
...
+            {
+                char json[] = "{\"req\": \"hub.sync\"}";
+
+                THEN("A CRC error SHALL NOT be reported")
+                {
+                    CHECK(!_crcError(json, seqNo));
+                }
+            }
+
+            AND_GIVEN("The CRC field exists, but is not at the tail of the response")
```
Bucknalla, line 130: What is the difference between this test and line #73? Are they expected to report differently?

zfields: The highest order branch (i.e. `GIVEN`), dictates whether `note-c` believes the Notecard firmware is capable of sending CRC's in the response.

- If the firmware does not support CRC, then a CRC is assumed to be of user origin and should not be tested.
- If the firmware does support CRC, but it's in the wrong position (then we assume it is of user origin and do not test it).

PR #171, opened 2025-01-30
fix for `_n_htoa32`

m-mcgowan:
fix: remove unused conditional define, which was causing the build to fail on 64-bit systems with `NOTE_C_SHOW_MALLOC` left undefined.  ~The `_n_htoa32` function isn't used unless `NOTE_C_SHOW_MALLOC` is defined, but it was being compiled because of the `!NOTE_C_LOW_MEM` part of the conditional, which evaluates to true by default.~

EDIT: Had to rework the code further - `_n_htoa32` is used in `n_request.c` when `NOTE_C_LOW_MEM` is not defined.  Moved the pointer to 32-bit ascii conversion to its own function, so that particular use case doesn't impact usage outside of reporting memory activity (`NOTE_C_SHOW_MALLOC`).

### DIFF
--- a/n_hooks.c
+++ b/n_hooks.c
@@ -520,7 +520,7 @@ void NoteDelayMs(uint32_t ms)
     }
 }
 
-#if NOTE_C_SHOW_MALLOC || !defined(NOTE_C_LOW_MEM)
+#if NOTE_C_SHOW_MALLOC
 //**************************************************************************/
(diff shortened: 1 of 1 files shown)


zfields:
Branch migrated to https://github.com/blues/note-c/pull/176, cleaned and merged.

PR #173, opened 2025-02-01
fix: static analysis issues

zfields:
- use comments to explain why these warnings are not valid
- suppress possible `NULL` pointer deference from `cppcheck`


rayozzie (changes requested):
AFAIK all of these changes are unnecessary and they are not adding any new protection.

PR #189, opened 2025-05-09
NF-1163: Update Notecard firmware nightly for card.binary HIL tests.

haydenroche5:
Other changes:

- Update workflow to use new notestation CLI.
- Add AWS credentials step to workflow. Needed in order to download Notecard firmware from S3.
- Run the tests nightly instead of weekly.


zfields (approved):
¯\_(ツ)_/¯

PR #196, opened 2025-06-12
TS-711: Fix notecard-binary-tests.yml missing inputs section.

haydenroche5:


haydenroche5:
The HIL failure is from a temporary ngrok failure. I re-ran it, and it succeeded: https://github.com/blues/note-c/actions/runs/15607213699

zfields (dismissed):

PR #201, opened 2025-08-26
chore: handle zero length request

zfields:
To enable "receive only" transactions (i.e `{heartbeat}`), I needed to first ensure zero length transmissions do not cause failures. Then I also attempted to homogenize `_i2cNoteTransaction()` and `_serialNoteTransaction()`.


Bucknalla (approved):
LGTM

PR #210, opened 2025-09-24
Improve cross-compilation support

jpetrina:
Hello,

one of our customers has an non-x86 Linux target and we needed to change a few things in the build system to make this library integrate well with e.g. firmware build systems like Buildroot and to support cross-compilation.

For example, it is great help for us to do:

```
-DCMAKE_SYSTEM_NAME=Linux
-DCMAKE_SYSTEM_PROCESSOR=arm
-DCMAKE_C_COMPILER=/usr/bin/arm-linux-gnueabihf-gcc
-DCMAKE_CXX_COMPILER=/usr/bin/arm-linux-gnueabihf-g++
-DCMAKE_FIND_ROOT_PATH=/opt/arm-sysroot
```

and it Just Works™.

I'm using this PR as an opportunity to open the discussion and contribute upstream; let me know if you'd like some changes, thanks!

cc @spahijajakov @PL4typus


zfields:
Thanks for bringing this up!

I just saw this, and haven't had a chance to properly review it.

I definitely want to support as many folks as possible, and I'm happy to introduce the flexibility required to make that happen. Let me grok what you've done, and I'll provide some feedback.

PR #225, opened 2026-02-23
build: fix cross-platform build support and add macOS CI

m-mcgowan:
## Summary

- Use `check_symbol_exists` to detect system-provided `strlcpy`/`strlcat` (macOS, FreeBSD, OpenBSD, glibc 2.38+) instead of bundling them unconditionally
- Guard Linux x86 32-bit cross-compile flags (`-m32`, `-mfpmath=sse`, linker paths) by `CMAKE_SYSTEM_PROCESSOR` so builds work on any POSIX platform
- Fix AppleClang warnings-as-errors at source (`NoteUserAgent` strict prototype, unused variables in `n_helpers.c`)
- Add a macOS build verification job to the CI pipeline
- Add `strlcpy`/`strlcat` to the libc dependency whitelist and make the lcov `n_str.c` exclude conditional

## Changes

**`CMakeLists.txt`:**
- `check_symbol_exists` for `strlcpy`/`strlcat`; define `HAVE_STRLCPY`/`HAVE_STRLCAT` and exclude `n_str.c` when the system provides them
- Guard `-m32`, `-mfpmath=sse`, `-msse2`, 32-bit linker paths, and `-Wl,-melf_i386` behind `CMAKE_SYSTEM_NAME`/`CMAKE_SYSTEM_PROCESSOR` checks (Linux x86 only)
- Add `-Wl,-flat_namespace` on Apple for FFF symbol interposition

**`test/CMakeLists.txt`:**
- Guard Catch2 `-m32` flag by Linux x86 check
- Add `-Wl,-flat_namespace` to test executables on Apple
- Only exclude `n_str.c` from lcov coverage when the file is actually compiled

**`n_ua.c` / `n_helpers.c`:**
- Fix `-Wstrict-prototypes`: `NoteUserAgent()` → `NoteUserAgent(void)`
- Remove unused variables (`j`, `lastLengthCount`) that triggered `-Wunused-but-set-variable`

**`scripts/check_libc_dependencies.sh`:**
- Add `strlcpy` and `strlcat` to the permitted libc function whitelist

**`.github/workflows/ci.yml`:**
- Add `run_macos_unit_tests` job (build-only; FFF interposition limitations prevent running ctest on macOS)

## Test plan

- [x] All existing unit tests build and pass on macOS (Apple Silicon)
- [x] note-posix builds and passes 13/13 tests using this branch
- [x] macOS CI job passes via `act` locally
- [x] Linux CI passes (verify after push of libc whitelist/lcov fixes)

PR #236, opened 2026-03-09
build: add reusable CMake library target for downstream projects

m-mcgowan:
Adds a reusable CMake library target, so that downstream targets don't have to deal with note-c internals, and can opt-in to build the tests.

This builds on PR #225,  since those fixes are needed to get a clean build on macOS.

### Summary                                                                                                                                                  

  - Adds a `note_c_lib` static library target that downstream projects (e.g. note-posix) can consume via `add_subdirectory` and `target_link_libraries`, without pulling in the unit test harness, 32-bit cross-compile flags, or FFF interposition settings, which are specific to the tests.
  - Defaults `NOTE_C_BUILD_TESTS` to ON only when note-c is the top-level project (using `PROJECT_IS_TOP_LEVEL`), so subdirectory consumers skip tests automatically.
  - Bumps `cmake_minimum_required` from 3.20 to 3.21 for` PROJECT_IS_TOP_LEVEL` support.
  - Includes an integration test (test/integration/) that exercises the add_subdirectory workflow end-to-end.

### Usage

```cmake
  add_subdirectory(note-c)
  target_link_libraries(my_app PRIVATE note_c_lib)
```
  The `note_c_lib` target carries the public include path, platform compile definitions (`HAVE_STRLCPY`, etc.), and all build options (`NOTE_C_LOW_MEM`, `NOTE_C_SINGLE_PRECISION`, etc.) — but none of the test-specific settings.

###  Test plan

  - [x] Existing unit tests pass (Linux CI unchanged)
  - [x] Integration test builds and passes (test/integration/)
  - [x] Verified with `https://github.com/zakoverflow/note-posix` — pristine clone builds and passes all 13 tests on macOS using `note_c_lib`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,6 +195,30 @@ jobs:
         run: |
           docker run --rm --volume $(pwd):/note-c/ --workdir /note-c/ --entrypoint ./scripts/run_cppcheck.sh ghcr.io/blues/note_c_ci:latest
 
+  run_macos_unit_tests:
+    runs-on: macos-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Configure
+        run: cmake -B build -DNOTE_C_BUILD_TESTS=ON
+
+      - name: Build
+        run: cmake --build build --parallel $(sysctl -n hw.ncpu)
+
+      # NOTE: ctest is not run here. A subset of tests that rely on FFF
+      # symbol interposition fail on macOS due to two-level namespace
+      # linking. The build step above verifies that all sources and tests
+      # compile and link successfully.
+      #
+      # Even with -Wl,-flat_namespace, intra-library calls within the
+      # note_c dylib are resolved at link time as direct calls, so FFF
+      # fakes in the test executable cannot interpose them. Fully fixing
+      # this would require linking note-c as a static library for tests
+      # or using a different mocking approach on macOS.
+
   publish_ci_image:
     runs-on: ubuntu-latest
     # Make sure unit tests unit tests passed before publishing.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,22 @@ jobs:
         run: |
           docker run --rm --volume $(pwd):/note-c/ --workdir /note-c/ --entrypoint ./scripts/check_libc_dependencies.sh ghcr.io/blues/note_c_ci:latest
 
+  run_integration_tests:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Configure add_subdirectory integration test
+        run: cmake -S test/integration -B build-integration
+
+      - name: Build add_subdirectory integration test
+        run: cmake --build build-integration --parallel $(nproc)
+
+      - name: Run add_subdirectory integration test
+        run: ctest --test-dir build-integration --output-on-failure
+
   run_unit_tests:
     runs-on: ubuntu-latest
     if: ${{ always() }}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.20)
+cmake_minimum_required(VERSION 3.21)
 cmake_policy(SET CMP0095 NEW)
 
 if ("${CMAKE_BINARY_DIR}" STREQUAL "${CMAKE_SOURCE_DIR}")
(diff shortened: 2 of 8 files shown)


m-mcgowan:
> Pretty cool! 🌟
> 
> I'm tending a dumpster fire at the moment 🔥, but I am excited to check this one out! 😎
> 
> Cheers!

Any updates?

zakoverflow:
Suggested cleanup around `NOTE_C_NO_LIBC`: I think this option should stay scoped to the shared `note_c` test/audit target, since its mechanism is link-time undefined-symbol auditing for `scripts/check_libc_dependencies.sh`.

The `note_c_lib` target is static, so its `PRIVATE` link options do not perform that audit at archive creation time, and they also should not propagate into downstream applications. Removing the block from `note_c_lib` makes that intent clearer while preserving the current CI check on `note_c`.

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 2eaa1a7..xxxxxxx 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,11 +21,18 @@ option(NOTE_C_LOW_MEM "Build the library tailored for low memory usage." OFF)
 option(NOTE_C_MEM_CHECK "Run tests with Valgrind." OFF)
 option(NOTE_NODEBUG "Build the library without debug information." OFF)
-option(NOTE_C_NO_LIBC "Build the library without linking against libc, generating errors for any undefined symbols." OFF)
+option(NOTE_C_NO_LIBC "Audit libc dependencies by linking the test shared library without libc." OFF)
 option(NOTE_C_SHOW_MALLOC "Build the library with flags required to log memory usage." OFF)
 option(NOTE_C_SINGLE_PRECISION "Use single precision for JSON floating point numbers." OFF)
 option(NOTE_C_HEARTBEAT_CALLBACK "Enable heartbeat callback support." OFF)
 
+if(NOTE_C_NO_LIBC AND NOT NOTE_C_BUILD_TESTS)
+    message(FATAL_ERROR
+        "NOTE_C_NO_LIBC is only meaningful for the note_c shared test/audit "
+        "target. Enable NOTE_C_BUILD_TESTS, or leave NOTE_C_NO_LIBC disabled "
+        "when consuming note_c_lib."
+    )
+endif()
+
 set(NOTE_C_SRC_DIR ${CMAKE_CURRENT_SOURCE_DIR})
@@ -97,15 +104,6 @@ else()
     )
 endif()
 
-if(NOTE_C_NO_LIBC)
-    target_link_options(
-        note_c_lib
-        PRIVATE
-            -nostdlib
-            -nodefaultlibs
-            LINKER:--no-undefined
-    )
-endif()
-
 if(NOTE_NODEBUG)
     target_compile_definitions(
         note_c_lib
```

The existing `NOTE_C_NO_LIBC` block on the shared `note_c` target should remain as-is, because that is the target the libc dependency scan actually validates.

zakoverflow:
I would also wire the new `test/integration` project into CI so the `add_subdirectory(note-c)` + `target_link_libraries(... note_c_lib)` contract is actually exercised on every PR.

Suggested workflow addition:

```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index c845f41..xxxxxxx 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,6 +198,22 @@ jobs:
         run: |
           docker run --rm --volume $(pwd):/note-c/ --workdir /note-c/ --entrypoint ./scripts/run_cppcheck.sh ghcr.io/blues/note_c_ci:latest
 
+  run_integration_tests:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Configure add_subdirectory integration test
+        run: cmake -S test/integration -B build-integration
+
+      - name: Build add_subdirectory integration test
+        run: cmake --build build-integration --parallel $(nproc)
+
+      - name: Run add_subdirectory integration test
+        run: ctest --test-dir build-integration --output-on-failure
+
   run_macos_unit_tests:
     runs-on: macos-latest
 
```

This catches regressions in the exact downstream-consumer workflow introduced by this PR without pulling in the unit-test harness, 32-bit test flags, or FFF-specific settings.

m-mcgowan:
Added the first commit to address your valid feedback, and a second commit as some cleanup to DRY up some definitions.

zfields (approved):
Thanks Mat! This was long overdue. 🙏

m-mcgowan:
> Thanks Mat! This was long overdue. 🙏

You're welcome.  I have an open issue from March (https://github.com/blues/note-c/issues/238) - keen to hear your thoughts on that perhaps with Ray's input? It's not specifically a note-c issue, more a notecard change, but note-c was the closest available place where I could raise the issue.

